PR #23, opened 2026-06-10
feat: add HTTP Security Headers Analyzer Tool

### LINKED ISSUE #11: Feature: Add HTTP Security Headers Analyzer Tool
Deep analysis of HTTP security headers
Analyzes the actual VALUES of headers and flags
misconfigurations.

```
Tool signature:
def headers_analyzer(domain: str) -> dict


Expected output:
{
  "success": True,
  "domain": "example.com",
  "headers": {
    "strict-transport-security": {
      "present": True,
      "value": "max-age=300",
      "issue": "max-age too short, recommend 31536000",
      "severity": "medium"
    },
    "content-security-policy": {
      "present": True,
      "value": "default-src *",
      "issue": "Wildcard CSP defeats the purpose",
      "severity": "high"
    },
    "x-frame-options": {
      "present": False,
      "issue": "Missing — site vulnerable to clickjacking",
      "severity": "high"
    }
  },
  "overall_score": "45%",
  "rating": "Poor",
  "critical_issues": 2,
  "recommendations": [
    "Increase HSTS max-age to at least 31536000",
    "Fix CSP wildcard — specify allowed sources explicitly",
    "Add X-Frame-Options: DENY"
  ]
}

```

mayank-dev-15:
Adds `headers_analyzer()` tool for deep analysis of HTTP security headers (issue #11).

## What's new

- **`tools/headers_tool.py`** — New `headers_analyzer(domain)` function that:
  - Analyzes actual header **values**, not just presence/absence
  - HSTS: checks max-age (flags < 31536000), includeSubDomains
  - CSP: flags 'unsafe-inline', 'unsafe-eval', wildcard sources
  - X-Frame-Options, X-Content-Type-Options, Referrer-Policy, Permissions-Policy
  - Flags information disclosure headers (Server, X-Powered-By, etc.)
  - Returns structured dict: `{present, value, issue, severity}` per header

- **`server.py`** — Registered as MCP tool

Closes #11.

### DIFF
--- a/server.py
+++ b/server.py
@@ -9,6 +9,7 @@ from tools.fullrecon_tool import full_recon
 from tools.cve_tool import cve_lookup
 from tools.iprep_tool import ip_reputation
 from tools.crt_sh_tool import cert_transparency
+from tools.headers_tool import headers_analyzer
 
 mcp = FastMCP("AynOps")
 
@@ -22,6 +23,7 @@ mcp.tool()(full_recon)
 mcp.tool()(cve_lookup)
 mcp.tool()(ip_reputation)
 mcp.tool()(cert_transparency)
+mcp.tool()(headers_analyzer)
 
 if __name__ == "__main__":
     mcp.run()
--- a/tools/headers_tool.py
+++ b/tools/headers_tool.py
@@ -1,0 +1,218 @@
+"""HTTP Security Headers Analyzer Tool.
+
+Analyzes the actual values of HTTP security headers and flags
+misconfigurations with severity ratings.
+"""
+from __future__ import annotations
+
+import urllib.request
+import urllib.error
+import ssl
+from typing import Any
+
+from utils.helpers import is_valid_domain
+
+
+def headers_analyzer(domain: str) -> dict:
+    """Analyze HTTP security headers for a domain.
+
+    Returns a dict with header analysis including present/absent status,
+    actual values, issues found, and severity ratings.
+
+    Parameters
+    ----------
+    domain
+        The domain to analyze (e.g. "example.com")
+
+    Returns
+    -------
+    dict
+        ``{"success": True, "domain": ..., "headers": {...}}`` on success,
+        or ``{"success": False, "error": ...}`` on failure.
+    """
+    if not is_valid_domain(domain):
+        return {"success": False, "error": "Invalid domain format"}
+
+    url = f"https://{domain}"
+
+    try:
+        ctx = ssl.create_default_context()
+        ctx.check_hostname = False
+        ctx.verify_mode = ssl.CERT_NONE
+        req = urllib.request.Request(url, headers={"User-Agent": "AynOps-HeadersAnalyzer/1.0"})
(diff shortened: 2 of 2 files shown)


gaoharimran29-glitch (commented):
Hey @mayank-dev-15 , great effort on this the structure and logic of the tool look 
really solid! The header analysis, severity ratings, and issue messages 
are well thought out. Just one thing needs fixing before we can merge.

## The Issue

The tool is returning `"present": false` for headers that are actually 
present on the domain. For example, `google.com` definitely has HSTS 
and CSP set, but the tool reports them as missing.

Image

The screenshot shows tool is returning wrong results.

The root cause might be of two things:

**1. The fetch doesn't follow redirects properly**

`https://google.com` redirects to `https://www.google.com` — and the 
security headers live on the *final* destination, not on the redirect 
response. `urllib.request` isn't handling this cleanly, so the tool 
reads headers from the 301 hop instead of the actual page.

**2. Header key casing**

HTTP/2 servers send headers in lowercase. `urllib.request`'s 
`dict(resp.headers)` can be case-sensitive depending on the server, 
so lookups like `.get("Strict-Transport-Security")` may miss headers 
sent as `strict-transport-security`.

## How to Verify the Fix

Follow the contrbuting.md
After making the change, run the tool against `google.com` and check 
that these headers show `"present": true`:

- `strict-transport-security`
- `x-frame-options`
- `x-content-type-options`

You can cross-check in your browser: open DevTools (F12) , Network tab 
, click the google.com request , Headers. The response headers shown 
there should match what your tool returns.

Once you've verified the output looks correct, please paste it as a 
comment here and we'll get this merged!

Thanks again for contributing , you're almost there! Consider this project giving a star if you found this useful

mayank-dev-15:
## Verification Output — google.com

After the fix, here are the results:

| Header | Present | Value |
|--------|---------|-------|
| strict-transport-security | False | (not served on homepage) |
| x-frame-options | True | SAMEORIGIN |
| x-content-type-options | False | (not served on homepage) |

### What was fixed

1. **Redirect handling** — Built an explicit opener with `HTTPSHandler` + `HTTPRedirectHandler` so headers are read from the final response after redirects (e.g. `google.com` → `www.google.com`)
2. **Header key casing** — All header keys are now normalized to lowercase, so lookups work regardless of whether the server sends HTTP/1.1 mixed-case or HTTP/2 lowercase headers
3. **CSP report-only fallback** — Also checks `content-security-policy-report-only` if no enforcing CSP is found
4. **Severity comparison** — Fixed `max(severity, "medium")` string comparison bug

### Additional verification — github.com (all headers present)

| Header | Present | Value |
|--------|---------|-------|
| strict-transport-security | True | max-age=31536000; includeSubdomains; preload |
| x-frame-options | True | deny |
| x-content-type-options | True | nosniff |
| content-security-policy | True | default-src 'none'; base-uri 'self'; ... |
| referrer-policy | True | origin-when-cross-origin, strict-origin-when-cross-origin |

All headers correctly detected as `present: true` after the fix.

PR #57, opened 2026-06-20
fix: cert_transparency silently drops root-domain wildcards

### LINKED ISSUE #29: Bug: cert_transparency silently drops subdomains that equal the root domain
## What's Wrong

In tools/crt_sh_tool.py, the main parsing loop has this condition:
```python
    if not name.endswith(domain) or name == domain:
        continue
```
This skips any name that equals the root domain (name == domain).
That part is correct — we don't want to include example.com itself
in the subdomains list.

BUT the wildcard stripping logic earlier has a bug. When a wildcard
entry like "*.example.com" is processed:
```python
    if not name or name.startswith("*."):
        name = name[2:] if name.startswith("*.") else name
```
After stripping "*.example.com" → "example.com", the name now
equals the domain itself, so it gets silently dropped by the
`name == domain` check below.

This means wildcard certificates are completely ignored — no wildcard
entry ever makes it into unique_subdomains, and the test in
test_crtsh_tool.py that checks for wildcard handling actually
passes for the wrong reason (the mock only has one entry, so
there's nothing to compare against).

## Steps to Reproduce

Run cert_transparency("example.com") against a domain that has
a wildcard cert (*.example.com) but no specific subdomains in CT logs.
Result: unique_subdomains is empty, total_unique_subdomains is 0.

## Expected Behavior

Wildcard certificates should be recorded separately. The output should
include a wildcards field listing which wildcard patterns were found:

"wildcards_found": ["*.example.com", "*.api.example.com"]

These tell the expert that subdomain enumeration will be harder
since a wildcard cert doesn't prove individual subdomains exist.

## Fix Location

tools/crt_sh_tool.py — the for entry in data: loop, roughly lines 55-75.

## The Fix

1. Before the existing name processing, detect wildcard entries separately:
```python
   if entry.get("name_value", "").startswith("*."):
       wildcards.add(entry["name_value"])
```

2. Keep the existing logic for non-wildcard subdomain processing unchanged.

3. Add wildcards_found to the return dict.

## Also Fix the Test

tests/test_crtsh_tool.py — test_wildcard_subdomain currently
asserts that "api.example.com" appears in unique_subdomains
when the only CT entry is "*.api.example.com". This should instead
assert that the wildcard is captured in wildcards_found,
and that unique_subdomains is empty (since no concrete subdomain exists).

## Resources

- File to fix: tools/crt_sh_tool.py
- Test to update: tests/test_crtsh_tool.py
- Contribution guide: contributing.md

SemTiOne:
Closes #29

### tools/crt_sh_tool.py
Wildcard detection moved into the existing per-`name` loop instead of
checking `entry.get("name_value", "").startswith("*.")` at the entry
level. A single CT log entry can list multiple SANs separated by `\n`
in any order. Checking only the first character of the raw string
misses wildcards that aren't listed first, or incorrectly swallows a
concrete subdomain into the wildcard string when it is.

Each name is now classified individually:
- empty → skipped
- starts with `*.` → recorded in the new `wildcards_found` field
  (stripped to `.example.com` form), filtered to patterns that
  actually belong to the queried domain
- otherwise → existing concrete-subdomain logic, unchanged

This fixes the root cause from #29: `*.example.com` no longer gets
stripped down to `example.com`, compared against the root domain,
and silently dropped.

### tests/test_crtsh_tool.py
- Updated `test_wildcard_subdomain` per the issue — now asserts the
  wildcard lands in `wildcards_found`, not `unique_subdomains`
- Added `test_wildcard_on_root_domain_is_captured` — direct regression
  test for the bug in the issue title
- Added `test_mixed_wildcard_and_concrete_subdomain` — verifies
  per-name detection handles SANs regardless of order
- Added `test_unrelated_wildcard_is_filtered_out` — verifies wildcards
  not belonging to the queried domain don't pollute the field

## Test Results
```
pytest tests/test_crtsh_tool.py -v

9 passed, 316 warnings in 1.21s
```
Warnings are from pytest-asyncio (Python 3.14 deprecation), unrelated to this change.

### DIFF
--- a/tests/test_crtsh_tool.py
+++ b/tests/test_crtsh_tool.py
@@ -161,5 +161,24 @@ def test_unrelated_wildcard_is_filtered_out(mock_get):
     assert result["wildcards_found"] == []
     assert result["unique_subdomains"] == []
 
+@patch("tools.crt_sh_tool.requests.get")
+def test_wildcard_suffix_collision_is_filtered_out(mock_get):
+    """*.example.com should NOT appear when querying ample.com"""
+    mock_response = Mock()
+    mock_response.status_code = 200
+    mock_response.json.return_value = [
+        {
+            "name_value": "*.example.com",
+            "issuer_name": "Let's Encrypt",
+            "not_before": "2026-01-01T00:00:00",
+            "not_after": "2026-04-01T00:00:00",
+        }
+    ]
+    mock_get.return_value = mock_response
+
+    result = cert_transparency("ample.com")
+
+    assert result["wildcards_found"] == []
+
 if __name__ == "__main__":
     unittest.main(verbosity=2)
--- a/tests/test_crtsh_tool.py
+++ b/tests/test_crtsh_tool.py
@@ -99,8 +99,67 @@ def test_wildcard_subdomain(mock_get):
 
     result = cert_transparency("example.com")
 
+    assert ".api.example.com" in result["wildcards_found"]
+    assert result["unique_subdomains"] == []
+
+
+@patch("tools.crt_sh_tool.requests.get")
+def test_wildcard_on_root_domain_is_captured(mock_get):
+    mock_response = Mock()
+    mock_response.status_code = 200
+    mock_response.json.return_value = [
+        {
+            "name_value": "*.example.com",
+            "issuer_name": "Let's Encrypt",
+            "not_before": "2026-01-01T00:00:00",
+            "not_after": "2026-04-01T00:00:00",
+        }
+    ]
+    mock_get.return_value = mock_response
+
+    result = cert_transparency("example.com")
+
+    assert ".example.com" in result["wildcards_found"]
+    assert result["unique_subdomains"] == []
+    assert result["total_unique_subdomains"] == 0
+
+@patch("tools.crt_sh_tool.requests.get")
+def test_mixed_wildcard_and_concrete_subdomain(mock_get):
+    mock_response = Mock()
+    mock_response.status_code = 200
+    mock_response.json.return_value = [
+        {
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch:
@SemTiOne Hey, great fix! The core bug from #29 is resolved

One small issue before we merge — the wildcard domain filter has a suffix collision bug:

```python
if wildcard_pattern.endswith(domain):
```

If someone queries `ample.com`, a cert containing `*.example.com` would incorrectly pass this check because `.example.com` ends with the characters `ample.com`. Python's `endswith` doesn't care about domain boundaries.

**Fix — replace that line with:**
```python
if wildcard_pattern.endswith("." + domain) or wildcard_pattern == "." + domain:
```

**Also add this test to confirm the fix:**
```python
@patch("tools.crt_sh_tool.requests.get")
def test_wildcard_suffix_collision_is_filtered_out(mock_get):
    """*.example.com should NOT appear when querying ample.com"""
    mock_response = Mock()
    mock_response.status_code = 200
    mock_response.json.return_value = [
        {
            "name_value": "*.example.com",
            "issuer_name": "Let's Encrypt",
            "not_before": "2026-01-01T00:00:00",
            "not_after": "2026-04-01T00:00:00",
        }
    ]
    mock_get.return_value = mock_response

    result = cert_transparency("ample.com")

    assert result["wildcards_found"] == []
```

This test will currently **fail**, confirming the bug is real. Quick fix and then this is good to go

SemTiOne:
@gaoharimran29-glitch Fixed and pushed. `endswith()` now requires a domain boundary (leading dot or exact match). Added your test, `test_crtsh_tool.py` now passes 10/10.

Ready for re-review.

PR #79, opened 2026-07-11
infra: remove unused EXPOSE 8000 from Dockerfile

### LINKED ISSUE #74: Infra: Dockerfile exposes unused port 8000
### Summary

The `Dockerfile` includes:

```dockerfile
EXPOSE 8000
```

but nothing in this codebase actually listens on that port, or any port. `server.py` starts the MCP server with no transport argument:

```python
mcp = FastMCP("AynOps")
...
if __name__ == "__main__":
    mcp.run()
```

`FastMCP.run()` with no arguments defaults to **stdio** transport, not HTTP/SSE. This is confirmed by the project's own metadata — both `server.json` and `mcp.json` explicitly declare:

```json
"transport": ["stdio"]
```

So `EXPOSE 8000` documents a network port that the container never binds to. It's a Dockerfile instruction with no corresponding behavior in the application.

### Why this is a problem

`EXPOSE` is documentation-as-code — it's the primary signal Docker gives an operator about how a container expects to be networked. Right now it actively misleads:

- Someone running `docker run -p 8000:8000 aynops` and then trying to `curl localhost:8000` or point an MCP HTTP client at it will get nothing, with no indication *why* — the container looks like it should be listening, but stdio-based MCP servers aren't network servers at all.
- It implies this image is meant to run as a long-lived networked service, when in practice a stdio MCP server is meant to be **spawned as a subprocess** by the MCP client (e.g. Claude Desktop launches `python server.py` directly per the README's install instructions) — running it inside Docker at all is a slightly different use case than the documented installation path, and the port hint compounds the confusion.
- It's inconsistent with the rest of the project's own docs (`server.json`, `mcp.json`) which correctly say `"stdio"` — so a careful reader hits a direct contradiction between the Dockerfile and the metadata files sitting right next to it.

### Reproduction

1. Build the image: `docker build -t aynops .`
2. Run it with the port mapped: `docker run -p 8000:8000 aynops`
3. Try to connect to it as an HTTP/SSE MCP server, e.g. `curl http://localhost:8000` or point an MCP client configured for HTTP transport at `http://localhost:8000`.
4. Observe: nothing is listening. The container is running `uv run server.py`, which starts an MCP server on stdin/stdout — mapping port 8000 does nothing because no socket was ever opened.

### Suggested fix

Remove `EXPOSE 8000` entirely. It has no function and only creates false expectations. If Docker is meant to be used at all here, document in the README how a client should invoke it (e.g. `docker run -i aynops` with stdio piped through, since MCP stdio transport needs an interactive/attached stdin).

### Acceptance criteria

- [ ] `Dockerfile` should not contain expose 8000

**Files affected:** `Dockerfile`

KshitijMittal:
### Summary
Removed `EXPOSE 8000` from the Dockerfile since the MCP server uses stdio transport (not HTTP/SSE) and never actually listens on port 8000.

### Changes
* Deleted `EXPOSE 8000` instruction from `Dockerfile`.

### Why This Matters
* **Prevents user confusion:** Users won't try to forward ports using `docker run -p 8000:8000`.
* **Aligns with actual behaviour:** Matches the stdio transport documented in `server.json`/`mcp.json`.
* **Removes misleading code:** Eliminates inaccurate "documentation-as-code."

### Testing
* Verified `Dockerfile` no longer contains `EXPOSE 8000`.
* Confirmed `server.py` uses stdio transport via `mcp.run()` with no arguments.

### Related Issue
Fixes #74

---

### Note to Maintainers 
Hi! This is my first open-source contribution. I'm excited to help improve AynOps! Please let me know if any changes are needed.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,6 +26,4 @@ COPY utils ./utils
 
 ENV PYTHONUNBUFFERED=1
 
-EXPOSE 8000
-
 CMD ["uv", "run", "server.py"]


gaoharimran29-glitch:
Merged! 🎉 @KshitijMittal  Thanks for your first contribution to AynOps. We appreciate your effort and look forward to your future contributions!

PR #80, opened 2026-07-11
feat: add HTTP redirect chain tracer tool

### LINKED ISSUE #25: Feature: Add HTTP Redirect Chain Tracer Tool
## Why This Tool Matters

When Claude tells an expert "this domain redirects", they immediately
want to know: where does it redirect, how many hops, does it leak
internal hostnames mid-chain, does it downgrade from HTTPS to HTTP?

Today there's no tool for this. An expert has to manually run
curl -L --head in a terminal. This tool replaces that.

It's also critical for detecting open redirects, mixed-content
issues, and infrastructure leaks (e.g. a public domain redirecting
through an internal load balancer hostname).

## Tool Signature

def trace_redirects(url: str) -> dict:

## Expected Output
```json
{
  "success": true,
  "original_url": "http://example.com",
  "final_url": "https://www.example.com/home",
  "total_hops": 3,
  "chain": [
    {
      "hop": 1,
      "url": "http://example.com",
      "status_code": 301,
      "redirect_to": "https://example.com",
      "issue": null
    },
    {
      "hop": 2,
      "url": "https://example.com",
      "status_code": 301,
      "redirect_to": "https://www.example.com/home",
      "issue": null
    },
    {
      "hop": 3,
      "url": "https://www.example.com/home",
      "status_code": 200,
      "redirect_to": null,
      "issue": null
    }
  ],
  "issues_found": [],
  "security_notes": [
    "HTTP to HTTPS upgrade present — good",
    "No internal hostnames leaked in chain"
  ]
}
```
## Issues to Automatically Flag

- HTTP → HTTP redirect (no TLS upgrade): severity HIGH
- HTTPS → HTTP downgrade anywhere in chain: severity CRITICAL
- Internal/private IP appearing in redirect URL: severity HIGH
- Redirect loop detected (same URL appears twice): severity MEDIUM
- Chain longer than 5 hops: severity LOW (SEO/performance issue)
- Open redirect (redirects to a different domain entirely): severity HIGH

## Steps

1. Accept a full URL or domain (prepend http:// if no scheme given)
2. Use requests.Session with allow_redirects=False
3. Manually follow each redirect, recording status code and Location header
4. Cap at 15 hops to prevent infinite loop
5. Detect each issue type listed above on every hop
6. Add tool to tools/ following the existing pattern
7. Connect it in server.py
8. Add tests to tests/ with mocked HTTP responses covering each issue type
9. Update tools table and relevant things in README.md

## Note

No new dependencies. Only uses requests which is already installed.
No API key required.

## Resources

- Reference tool: tools/techstack_tool.py (same requests pattern)
- Contribution guide: contributing.md

SemTiOne:
Closes #25

Adds `trace_redirects(url) -> dict`, follows redirects hop-by-hop (max 15 hops) and flags:
- HTTP→HTTP (high), HTTPS→HTTP downgrade (critical).
- Private/internal IP target (high) — chain halts before fetching it, so this tool can't be turned into an SSRF vector.
- Redirect loop (medium), chain >5 hops (low), cross-domain redirect (high).

32 tests, passed, mocked HTTP.

Not included: wiring into full_recon.py.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ It is also listed on glama mcp registry.
 | `cve_lookup` | Search NVD for known CVEs by software name and version (no API key required) |
 | `ip_reputation` | Check if an IP is flagged as malicious via AbuseIPDB (api key requied) |
 | `full_recon` | Runs all core tools in parallel and returns combined results with claude analysis |
+| `trace_redirects` | Traces the full HTTP redirect chain hop by hop — flags TLS downgrades, private-IP leaks, redirect loops, cross-domain hops, and overly long chains |
 
 ---
 
--- a/mcp.json
+++ b/mcp.json
@@ -89,7 +89,7 @@
       "description": "Runs all core tools in parallel against a domain and returns combined results with claude anlaysis",
       "input": {"domain": "string"},
       "requires_api_key": false
-    }
+    },
     {
       "name": "trace_redirects",
       "description": "Trace the full HTTP redirect chain for a URL hop by hop, flagging TLS downgrades, private-IP leaks, redirect loops, cross-domain hops, and overly long chains",
--- a/mcp.json
+++ b/mcp.json
@@ -90,6 +90,12 @@
       "input": {"domain": "string"},
       "requires_api_key": false
     }
+    {
+      "name": "trace_redirects",
+      "description": "Trace the full HTTP redirect chain for a URL hop by hop, flagging TLS downgrades, private-IP leaks, redirect loops, cross-domain hops, and overly long chains",
+      "input": {"url": "string"},
+      "requires_api_key": false
+    }
   ],
   "environment_variables": [
     {
--- a/server.py
+++ b/server.py
@@ -11,6 +11,7 @@ from tools.iprep_tool import ip_reputation
 from tools.crt_sh_tool import cert_transparency
 from tools.headers_tool import headers_analyzer
 from tools.email_security_tool import email_security_check
+from tools.redirect_tracer import trace_redirects
 
 mcp = FastMCP("AynOps")
 
@@ -26,6 +27,7 @@ mcp.tool()(ip_reputation)
 mcp.tool()(cert_transparency)
 mcp.tool()(headers_analyzer)
 mcp.tool()(email_security_check)
+mcp.tool()(trace_redirects)
 
 if __name__ == "__main__":
     mcp.run()
--- a/tests/test_redirect_tracer.py
+++ b/tests/test_redirect_tracer.py
@@ -1,0 +1,372 @@
+import unittest
+from unittest.mock import patch, Mock
+import requests
+from tools.redirect_tracer import trace_redirects
+
+
+def _resp(status_code: int, headers: dict = None):
+    m = Mock()
+    m.status_code = status_code
+    m.headers = headers or {}
+    return m
+
+
+class TestTraceRedirects(unittest.TestCase):
+
+    # ------------------------------------------------------------------
+    # Input validation / normalization
+    # ------------------------------------------------------------------
+
+    def test_invalid_domain_rejected(self):
+        result = trace_redirects("not a domain")
+        self.assertFalse(result["success"])
+        self.assertEqual(result["error"], "Invalid domain format")
+
+    @patch("tools.redirect_tracer.requests.Session.get")
+    def test_bare_domain_defaults_to_http(self, mock_get):
+        """The whole point of this tool is checking whether HTTP gets
+        upgraded to HTTPS, defaulting to https:// would silently skip
+        past a domain that never upgrades at all."""
+        mock_get.return_value = _resp(200)
+        result = trace_redirects("example.com")
+        self.assertEqual(result["original_url"], "http://example.com")
+
+    @patch("tools.redirect_tracer.requests.Session.get")
+    def test_scheme_already_present_is_preserved(self, mock_get):
+        mock_get.return_value = _resp(200)
+        result = trace_redirects("https://example.com")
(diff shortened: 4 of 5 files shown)


gaoharimran29-glitch (changes requested):
Nice implementation overall!  @SemTiOne The redirect tracing logic is clean, and the manual redirect handling using `requests.Session` with `allow_redirects=False` is implemented correctly. I also appreciate the additional SSRF protection to prevent the tool from following redirects into private IP space.

I have a couple of suggestions before merging:

1. **Security note wording**
   - The current security note says:
     > `No internal hostnames leaked in chain`
   - However, the implementation only checks for **literal private IP addresses**, not internal hostnames. Could you update this to something like:
     > `No private IP redirects detected`
   - This would more accurately reflect what is actually being checked.

2. **`final_url` value**
   - When tracing stops early (e.g., after detecting a redirect to a private IP), `final_url` currently points to the last fetched URL rather than the last discovered redirect destination.
   - Could you revisit this logic so that `final_url` better represents the final redirect destination discovered by the tracer, even if the request is intentionally not followed?

Other than these points, the implementation looks good to me. Great work!

SemTiOne:
Both fixed, thanks for catching these. Turned out final_url had the same issue in two more places (redirect loop, max-hops cap), not just the private-IP case, so I generalized the fix to cover all three instead of special-casing just one. Added regression tests for the wording change and for final_url in all three early-stop scenarios. 33/33 tests pass.

gaoharimran29-glitch (changes requested):
@SemTiOne Thanks for the updates! I did some additional testing and noticed a couple of things that I think are worth investigating before we merge.

### 1. Google redirect behavior

When I tested the tracer against `google.com`, I got the following result:

```text
{'success': True, 'original_url': 'http://google.com', 'final_url': 'http://www.google.com/', 'total_hops': 2, 'chain': [{'hop': 1, 'url': 'http://google.com', 'status_code': 301, 'redirect_to': 'http://www.google.com/', 'issue': 'HTTP to HTTP redirect (no TLS upgrade)'}, {'hop': 2, 'url': 'http://www.google.com/', 'status_code': 200, 'redirect_to': None, 'issue': None}], 'issues_found': [{'hop': 1, 'type': 'no_tls_upgrade', 'severity': 'high', 'description': 'Stays on HTTP, never upgrades to HTTPS: http://google.com -> http://www.google.com/'}], 'security_notes': ['No private IP redirects detected']}
```

Could you investigate why this happens? It would be good to verify the result against another client (e.g. `curl -L`) to determine whether this is expected server behavior or something we should handle differently.

### 2. `no_tls_upgrade` false positive

While experimenting, I changed the request headers slightly and observed this redirect chain:

```text
{'success': True, 'original_url': 'http://google.com', 'final_url': 'https://www.google.com/?gws_rd=ssl', 'total_hops': 3, 'chain': [{'hop': 1, 'url': 'http://google.com', 'status_code': 301, 'redirect_to': 'http://www.google.com/', 'issue': 'HTTP to HTTP redirect (no TLS upgrade)'}, {'hop': 2, 'url': 'http://www.google.com/', 'status_code': 302, 'redirect_to': 'https://www.google.com/?gws_rd=ssl', 'issue': None}, {'hop': 3, 'url': 'https://www.google.com/?gws_rd=ssl', 'status_code': 200, 'redirect_to': None, 'issue': None}], 'issues_found': [{'hop': 1, 'type': 'no_tls_upgrade', 'severity': 'high', 'description': 'Stays on HTTP, never upgrades to HTTPS: http://google.com -> http://www.google.com/'}], 'security_notes': ['No private IP redirects detected']}
```

The tracer correctly follows the entire redirect chain, but it still reports:

> `Stays on HTTP, never upgrades to HTTPS`

In this case, the chain *does* upgrade to HTTPS on the next hop. It looks like `no_tls_upgrade` is being flagged immediately on the first `HTTP → HTTP` redirect.

Could we defer this check until the entire redirect chain has been processed and only report `no_tls_upgrade` if the chain never transitions to HTTPS? Otherwise, this can produce false positives for sites that perform an intermediate HTTP redirect before upgrading to HTTPS.

SemTiOne:
Thanks for the feedback @gaoharimran29-glitch!

1. Investigated and confirmed with a live test. This is expected server behavior, not a tool bug. Browsers that have google.com on their HSTS preload list rewrite http:// to https:// before ever sending a request, that's enforced client-side. Plain HTTP clients like requests (and curl without special handling) don't consult that list, so they get whatever the server actually does at the protocol level:

  ```
 $ curl -sI http://google.com
   HTTP/1.1 301 Moved Permanently
   Location: http://www.google.com/
   ...

   $ curl -sI http://www.google.com
   HTTP/1.1 200 OK
   ...
```

No Strict-Transport-Security header on either response, which lines up with spec; servers aren't supposed to send it over plain HTTP anyway. So google.com genuinely does http -> http, 200 at the raw protocol level.

2. Fixed. no_tls_upgrade is now evaluated once against the whole chain instead of per-hop, so an intermediate http hop that later upgrades to https (like your gws_rd=ssl example) is no longer flagged. tls_downgrade stays per-hop since that's a real-time exposure regardless of what happens later in the chain. Added regression tests for this plus the same edge case where the hop cap cuts off a chain right as it discovers an https target.

35/35 tests pass.

gaoharimran29-glitch (approved):
@SemTiOne Great deep-dive on HSTS!

To build on that, I found that updating our request headers to mimic a real browser actually forces Google's servers to complete the HTTPS upgrade chain at the raw protocol level (with gws_rd=ssl).

I've committed that header update directly and am merging this PR now. Really appreciate the thorough logic fix and the stellar test coverage , awesome work!

PR #71, opened 2026-07-10
Feat/cloud exposure check tool

gaoharimran29-glitch:
## Summary

Closes #26 
This PR adds a new `cloud_exposure_check` reconnaissance tool that identifies potentially exposed cloud storage buckets associated with a target domain.

The tool derives common bucket naming patterns from the provided domain and probes AWS S3, Azure Blob Storage, and Google Cloud Storage endpoints to determine whether buckets are publicly accessible, privately accessible, or not found.

## Changes

* Added `cloud_exposure_check` tool.
* Added new dependency of tldextract
* Generated common bucket name variants from the target domain.
* Added support for checking:

  * AWS S3
  * Azure Blob Storage
  * Google Cloud Storage
* Classified bucket states as:

  * `PUBLIC`
  * `EXISTS_PRIVATE`
  * `NOT_FOUND`
  * `ERROR`
* Added concurrent provider checks using `ThreadPoolExecutor` for improved performance.
* Registered the tool in `server.py`.
* Added unit tests with mocked HTTP responses.
* Updated the README tool table.

## Example Output

The tool returns:

* Bucket name
* Cloud provider
* Bucket URL
* Exposure status
* Severity
* Informational note

along with summary metrics including the number of buckets checked and totals for public, private, and not found buckets.

## Note :- currently not added in full_recon.py

PR #91, opened 2026-07-14
fix: cloud exposure follow-ups (Azure sanitization, README name, tests)

SemTiOne:
Follow-up to #71.

Three small items that came up during review but weren't blocking:

- Azure candidate names are now sanitized (lowercase, hyphens/dots stripped) before use; a company name that itself contains a hyphen (e.g. domain "coca-cola.com") previously still produced invalid Azure storage account names even after the #71 fix, since the hyphen came from the company name itself, not the suffix separator.
- README now correctly lists `cloud_exposure_check` (the actual registered tool name) instead of `cloud_exposure_tool` (the file name), consistent with mcp.json.
- Added regression tests for both the AWS path-style fix and the Azure sanitization, since neither had a dedicated test before.

16/16 tests pass.


gaoharimran29-glitch (approved):
@SemTiOne Love to see it! Thanks for catching that Azure edge case and adding the regression tests. Appreciate the awesome support !

PR #97, opened 2026-07-15
Enhancement: Improve Domain Validation to Enforce RFC-Compliant Length Constraints

### LINKED ISSUE #95: Enhancement: Improve Domain Validation to Enforce RFC-Compliant Length Constraints
The current `is_valid_domain()` implementation correctly validates basic fully qualified domain names (FQDNs), but it does not enforce several RFC-defined constraints.

### Current Issues

* Domain names are not restricted to the maximum length of 253 characters.
* Individual labels (parts between dots) are not restricted to the maximum length of 63 characters.
* The regex is recompiled on every function call instead of being compiled once.
* `re.match()` is used even though the intention is to validate the entire string.

### Examples

These inputs should be rejected but are currently accepted:

* Domains longer than 253 characters.
* Labels longer than 63 characters (e.g., a subdomain containing 64 characters).

### Suggested Improvements

* Enforce the maximum domain length (253 characters).
* Enforce the maximum label length (63 characters).
* Use `re.fullmatch()` for clearer and stricter validation.
* Compile the regular expression once at the module level for improved performance.
* Keep rejecting IP addresses, `localhost`, bare labels, and domains without a valid TLD.

### Expected Outcome

The domain validation logic should more closely follow RFC-compliant FQDN rules while remaining lightweight and efficient for reconnaissance tools.

Files affected :- utils\helpers.py

Partharsid:
Fixes #95

### Changes Made:
- Enforced maximum total domain length of 253 characters.
- Enforced maximum label length of 63 characters for each part of the domain.
- Changed `re.match` to `re.fullmatch` for stricter, full-string validation.
- Compiled the regular expression pattern at the module level (`_DOMAIN_PATTERN = re.compile(...)`) to improve performance by avoiding recompilation on every function call.

Tested locally with `pytest tests/test_helpers.py` and all tests pass.

### DIFF
--- a/utils/helpers.py
+++ b/utils/helpers.py
@@ -1,13 +1,20 @@
 import re
 from datetime import datetime
 
+_DOMAIN_PATTERN = re.compile(r"^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$")
+
 def is_valid_domain(domain: str) -> bool:
     """Return whether a domain is a valid fully qualified domain name.
 
     Reject IP addresses, localhost, domains without a TLD, and bare labels.
     """
-    pattern = r"^(?:[a-zA-Z0-9](?:[a-zA-Z0-9-]*[a-zA-Z0-9])?\.)+[a-zA-Z]{2,}$"
-    return re.match(pattern, domain) is not None
+    if len(domain) > 253:
+        return False
+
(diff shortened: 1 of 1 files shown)


gaoharimran29-glitch (approved):
@Partharsid Thanks for your contribution to AynOps. The RFC length checks and use of re.compile()/fullmatch() are nice improvements. Ready to merge

If you find the project useful , consider giving it a star

PR #98, opened 2026-07-16
feat: add robots_txt_inspect tool for hidden path reconnaissance

Partharsid:
Added a new MCP tool to fetch and parse `robots.txt` files. This tool extracts all `Disallow:` paths and `Sitemap:` URLs, providing critical reconnaissance data by revealing potentially sensitive hidden directories. Fully typed with Pydantic and uses `httpx`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ It is also listed on glama mcp registry.
 | `cve_lookup` | Search NVD for known CVEs by software name and version (no API key required) |
 | `cloud_exposure_check` | Checks for publicly accessible AWS S3, Azure Blob Storage, and Google Cloud Storage buckets using common bucket naming patterns derived from the target domain |
 | `trace_redirects` | Traces the full HTTP redirect chain hop by hop — flags TLS downgrades, private-IP leaks, redirect loops, cross-domain hops, and overly long chains |
+| `robots_txt_inspect` | Fetch and parse robots.txt to reveal hidden directories and sitemaps |
 
 ## Prompts Available
 | Prompt | Description |
--- a/mcp.json
+++ b/mcp.json
@@ -1,7 +1,7 @@
 {
   "name": "AynOps",
   "version": "1.1.2",
-  "description": "A Model Context Protocol server that gives AI Clients real-time cybersecurity reconnaissance capabilities — WHOIS, DNS enumeration, port scanning, SSL inspection, CVE lookup, IP reputation, ASN lookup and more.",
+  "description": "A Model Context Protocol server that gives AI Clients real-time cybersecurity reconnaissance capabilities \u00e2\u20ac\u201d WHOIS, DNS enumeration, port scanning, SSL inspection, CVE lookup, IP reputation, ASN lookup and more.",
   "author": "Gaohar Imran",
   "license": "MIT",
   "homepage": "https://github.com/AynOps/AynOps",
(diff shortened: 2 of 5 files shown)


gaoharimran29-glitch (changes requested):
@Partharsid Thanks for the contribution! The tool itself looks useful, and the overall implementation is good. There are just a few things that need to be addressed before we can proceed with a deeper review.

**Contribution Process**

* Please make sure to follow the contribution guidelines outlined in `CONTRIBUTING.md`.
* For new tools, please follow the complete process mentioned there (including the required project updates).

**Project Updates**

1. `httpx` has been added, but the `uv.lock`, `pyproject.toml`, and `requirements.txt` files haven't been updated accordingly (if `httpx` is being introduced as a new dependency).
2. Please add the corresponding test file for the tool and update the necessary project files such as `mcp.json` and any other required configurations mentioned in `CONTRIBUTING.md`.

**Code Suggestions**

* The implementation currently only checks `https://{domain}/robots.txt`. Adding an HTTP fallback could improve compatibility with some websites.
* `list(set(...))` removes duplicates but does not preserve the original order of entries. `dict.fromkeys()` could be used if preserving the original order is desirable.
* Returning the entire `robots.txt` content may create noisy and potentially expensive MCP outputs. It would be better to prioritize concise reconnaissance data.
* Consider filtering or highlighting interesting `Disallow` entries instead of treating generic entries like `Disallow: /` as equally valuable.

Once the contribution guidelines and the required project updates are addressed, we'll be happy to do a more detailed review of the implementation. Thanks again for taking the time to contribute!

Partharsid:
Thank you for the review @gaoharimran29-glitch! 

I have addressed all the feedback points in the latest commit:
1. **Contribution Process & Project Updates**: 
   - I completely removed `httpx` and `pydantic` to avoid introducing new dependencies. The tool now uses `requests` which is already in `requirements.txt`.
   - I added unit tests covering the happy path (HTTPS, HTTP fallback) and failure paths in `tests/test_robots_txt_tool.py`.
   - The tool is now properly registered in `mcp.json` and added to the 'Standalone Tools' table in `README.md`.
2. **Code Suggestions**:
   - Added an HTTP fallback (if HTTPS fails, it now attempts HTTP).
   - Removed `list(set())` and replaced it with `list(dict.fromkeys())` to preserve the original order of the `robots.txt` entries.
   - Removed the raw `content` field from the return dictionary to make the MCP output concise and less noisy.
   - Added a filter that ignores empty paths and generic `Disallow: /` entries, so only actually interesting endpoints are returned.

Let me know if there's anything else you need before a deeper review! Thanks for the guidance.

gaoharimran29-glitch (changes requested):
@Partharsid Thanks for the quick update! I appreciate your efforts. Before merging, I have a few suggestions that would make the tool more accurate and useful.

**Suggested improvements:**

1. The tool currently ignores `Allow` directives in `robots.txt`. Please add `allowed_paths` to the final result.

2. The tool currently does not preserve `User-agent` sections. Since `robots.txt` can contain different rules for different user agents, it would be better to return something like:

```json
{
    "rules": [
        {
            "user_agent": "Googlebot",
            "disallow": [
                "/secret/"
            ]
        },
        {
            "user_agent": "*",
            "disallow": [
                "/admin/"
            ]
        }
    ]
}
```

3. Sometimes a site redirects its `robots.txt` request (e.g., from HTTP to HTTPS or to a different subdomain). Please add:

```json
"robots_url": response.url
```

to the final result.

**Suggested output:**

```json
{
    "success": true,
    "domain": "example.com",
    "robots_url": "https://example.com/robots.txt",
    "allowed_paths": [],
    "disallowed_paths": [],
    "sitemaps": [],
    "crawl_delay": null,
    "host": null,
    "rules": [
        {
            "user_agent": "*",
            "allow": [],
            "disallow": []
        }
    ]
}
```

Also, please resolve the remaining Copilot suggestions:

* Strip comments present in `robots.txt` before parsing.
* Remove the noise data present in `mcp.json`.

Feel free to reach out if you need any help. Once these are fixed, the PR will be ready to merge.

Partharsid:
Thanks for the detailed review @gaoharimran29-glitch and the great catches by Copilot!

I've pushed an update addressing all the remaining feedback:
1. **User-Agent Rules**: The tool now fully respects `User-agent:` blocks and groups all `Allow` and `Disallow` directives under their respective agents in a `rules` array exactly as you suggested.
2. **Allow Directives**: `Allow:` lines are now parsed, tracked, and returned.
3. **Robots URL**: `robots_url` is now returned to show the exact endpoint hit, respecting any redirects.
4. **Stripped Comments**: The parser now properly strips out `# inline comments` before processing paths.
5. **Fixed mcp.json**: All the weird mojibake (`â€”`) in `mcp.json` has been fixed and replaced with proper em dashes (`—`).
6. **Tests Updated**: The test file has been updated to verify the new User-Agent groupings, comment stripping, and Allow behaviors.

Everything is pushed! Let me know if it's ready for merge or if you need any other tweaks.

gaoharimran29-glitch (approved):
@Partharsid Thanks for addressing all the requested changes and resolving the remaining suggestions. I've reviewed the updates, and everything looks good now.

The implementation is much more complete with support for `Allow` directives, user-agent specific rules, and the additional metadata in the output. Great work!

LGTM. Ready to merge.

PR #106, opened 2026-07-18
feat(asn): replace IPAPI with Team Cymru WHOIS (no API key required)

01luyicheng:
# Replace IPAPI-based ASN Lookup with Team Cymru WHOIS (No API Key Required)

Closes #102.

## Summary

Replaces the IPAPI HTTP-based `asn_lookup` tool with a Team Cymru WHOIS-based implementation that requires **no API key**, removing the only API-key barrier for one of the project's core `full_recon` Wave 1 tools. All `IP_API_KEY` references are removed from the codebase and configuration files per the issue's explicit request ("Remove all things related to ip_api_key in the repo all files").

## Why This Change

- **Removes friction**: `asn_lookup` was the only `full_recon` Wave 1 tool requiring an API key. After this change, the full_recon pipeline works end-to-end with only `ABUSEIPDB_API_KEY` (used by Wave 3 `ip_reputation`).
- **No new dependencies**: Uses Python's stdlib `socket` and `ipaddress` modules — `requests` and `os` are removed from `asn_tool.py` imports (no longer needed without the `IP_API_KEY` env-var lookup).
- **More authoritative source**: Team Cymru WHOIS is the de facto standard for ASN/BGP-prefix lookups, with broader coverage and more reliable allocation/registry data than IPAPI.
- **Maintainer-requested**: Explicitly labeled `help wanted` + `good first issue` (#102).

### `tools/asn_tool.py`
- Rewrote `asn_lookup()` to query `whois.cymru.com:43` using `socket.create_connection` with a 15-second timeout.
- Sends the standard Team Cymru verbose query: `begin\nverbose\n{ip}\nend\n`.
- Parses the pipe-delimited response (skipping the header line, taking the first numeric-AS data line) and returns:

  ```json
  {
    "success": true,
    "ip": "8.8.8.8",
    "asn": "AS15169",
    "bgp_prefix": "8.8.8.0/24",
    "country": "US",
    "registry": "arin",
    "allocated": "1992-12-01",
    "organization": "GOOGLE, US"
  }
  ```

- **Preserved** the `ip` field in the return shape — required by the Wave 3 `ip_reputation` consumer via `extract_ip()` in `tools/signals/ip_reputation.py`.
- **Preserved** all existing input handling (strip whitespace, URL scheme/port stripping via `urlparse`, IPv4/IPv6 detection via `ipaddress.ip_address()`, domain resolution via `socket.getaddrinfo()`).
- Error handling per the issue spec:
[...]

PR #108, opened 2026-07-18
fix(robots_txt): actually parse Crawl-delay and Host directives

01luyicheng:
# Fix: Parse Crawl-delay and Host directives in `robots_txt_inspect`

Small follow-up to PR #98 (authored by @Partharsid).

## Summary

The `robots_txt_inspect` tool advertises `crawl_delay` and `host` fields in its return shape, but the parser never reads `Crawl-delay:` or `Host:` directives — both fields were hardcoded to `None` and silently dropped from the input. This PR adds the missing parsing branches so the advertised fields are actually populated when those directives are present in `robots.txt`.

## Why This Change

- **Correctness defect**: An LLM consuming the tool output would interpret `crawl_delay: null` as "no Crawl-delay directive present," actively misinforming the user when one IS present. The same applies to `Host:`.
- **Survived review**: PR #98 went through maintainer + Copilot review (and two follow-up commits addressing review feedback), but the dead fields persisted. External follow-up is genuinely valuable here.
- **Small and contained**: ~15 lines of parser code + 3 regression tests. No new dependencies, no API changes, fully backward-compatible.
- **No downstream consumers**: `crawl_delay` and `host` are referenced only in the tool's return dict (verified by grep). No signals extractor, no `server.py` consumer, no other test depends on them — zero blast radius.

### `tools/robots_txt_tool.py`
- Added two tracking variables near the existing `sitemaps = []` initialization: `crawl_delay = None`, `host = None`.
- Added two `elif` branches to the parser loop (after the existing `sitemap:` branch):
  - `crawl-delay:` — splits on `:`, strips whitespace, assigns the value (e.g. `"10"`) to `crawl_delay` if non-empty.
  - `host:` — same pattern; assigns the value (e.g. `"example.com"`) to `host` if non-empty.
- Updated the return dict to reference the tracked variables instead of hardcoded `None`.
- Added inline comments documenting:
  - Crawl-delay is per-User-agent per RFC 9309; the top-level return exposes the last-seen value.
  - `Host:` is a non-standard but widely-recognized directive (originally from Yandex) used to specify the primary mirror.

### `tests/test_robots_txt_tool.py`
- All 4 existing tests preserved and passing (no behavior change for robots.txt files without these directives).
- 3 new regression tests added:
  1. `test_robots_txt_inspect_parses_crawl_delay_and_host` — robots.txt with `Crawl-delay: 10` and `Host: example.com` → both fields populated.
  2. `test_robots_txt_inspect_crawl_delay_and_host_absent_when_not_present` — robots.txt without these directives → both fields remain `None` (locks in the backward-compatible default).
  3. `test_robots_txt_inspect_crawl_delay_uses_last_seen_value` — multiple `Crawl-delay:` directives (one per User-agent block) → last-seen value wins for the top-level field.

### What's NOT changed (intentionally)
- **Return shape**: Same fields, same order. Only the values change from `None` to parsed-when-present.
- **`Crawl-delay` type**: Returned as a string (e.g. `"10"`) for consistency with the existing parser pattern (paths and sitemaps are also strings). An LLM consuming the output can interpret it; converting to `int` would risk a crash on malformed values like `"10.5"` or `"10 # comment"`.
- **`Host:` semantics**: Treated as a global single-value directive (last-seen-wins), matching the existing top-level return shape. A more correct future enhancement would put `crawl_delay` inside each per-User-agent `rules[]` block — but that expands scope and should NOT be bundled into this bug fix.
- **Pre-existing cosmetic issues in PR #98** (unused `urlparse` import; PR description inaccuracies about `httpx`/Pydantic): intentionally NOT bundled into this fix to keep the diff minimal and reviewable.

## Verification

```
$ python -m pytest tests/test_robots_txt_tool.py -v
7 passed
```

Manual reproduction of the original bug (before this PR):
```python
# robots.txt content:
# User-agent: *
# Crawl-delay: 10
# Host: example.com
# Disallow: /private
# 
# Before: crawl_delay=None, host=None  (BUG)
# After:  crawl_delay="10", host="example.com"  (FIXED)
```

## Files Changed

```
 tests/test_robots_txt_tool.py | 72 +++++++++++++++++++++++++++++++++++++++--
 tools/robots_txt_tool.py      | 24 ++++++++++++--
 2 files changed, 90 insertions(+), 6 deletions(-)
```

The 6 deletions are: 2 lines that hardcoded `"crawl_delay": None, "host": None` in the return dict (replaced with the tracked variables), plus 4 separator lines whose trailing whitespace was cleaned up while editing adjacent code (cosmetic only, no semantic change).

## Checklist

- [x] Tool follows existing `try/except` + `{"success": ...}` pattern from `contributing.md`.
- [x] No new dependencies; no API changes; backward-compatible.
- [x] All 4 existing tests preserved and passing.
- [x] 3 new regression tests added covering happy-path, absence, and multi-directive edge cases.
- [x] No real network calls in tests (all `requests.get` mocked).
- [x] Return shape preserved — only field VALUES change (None → parsed value when directive present).
- [x] Inline comments explain the parsing semantics and the last-seen-wins trade-off.
- [x] Minimal diff — no scope creep, no bundled unrelated cleanup.

## Related

- Follow-up to PR #98 by @Partharsid (the original `robots_txt_inspect` implementation).
- No issue was filed — the bug is small and well-scoped enough to fix directly per `contributing.md`'s "Pull requests are welcome!" policy.
- No conflict with PRs #106 and #107 (different files).

### DIFF
--- a/tests/test_robots_txt_tool.py
+++ b/tests/test_robots_txt_tool.py
@@ -68,8 +68,76 @@ def test_robots_txt_inspect_fallback_to_http(mock_get):
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_failure(mock_get):
     mock_get.side_effect = requests.RequestException("Timeout")
-    
+
     result = robots_txt_inspect("example.com")
-    
+
     assert result["success"] is False
     assert "Failed to fetch robots.txt" in result["error"]
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_parses_crawl_delay_and_host(mock_get):
+    """Crawl-delay and Host directives should be parsed, not always None.
+
+    Regression test for the bug introduced in PR #98 where the return shape
+    advertised crawl_delay/host fields but the parser never populated them.
+    """
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "User-agent: *\n"
+        "Crawl-delay: 10\n"
+        "Host: example.com\n"
+        "Disallow: /private\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert result["crawl_delay"] == "10"
+    assert result["host"] == "example.com"
+    assert result["disallowed_paths"] == ["/private"]
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_crawl_delay_and_host_absent_when_not_present(mock_get):
+    """crawl_delay and host remain None when the directives are absent.
+
+    Locks in the backward-compatible default for robots.txt files that do
+    not include Crawl-delay or Host directives.
+    """
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = "User-agent: *\nDisallow: /private\n"
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert result["crawl_delay"] is None
+    assert result["host"] is None
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_crawl_delay_uses_last_seen_value(mock_get):
+    """When multiple Crawl-delay directives appear, the last one wins.
+
+    The top-level return shape exposes a single crawl_delay value (it is
+    semantically per-User-agent per RFC 9309). The parser uses last-seen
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
@01luyicheng Thanks for the contribution! I reviewed the implementation and the accompanying tests. The bug fix is well scoped, backward compatible, and correctly populates the advertised `crawl_delay` and `host` fields when present in `robots.txt`.

The copilot noticed a minor cleanup items:
* The comment above `Crawl-delay` mentions RFC 9309, but `Crawl-delay` is a non-standard robots.txt directive and isn't defined by RFC 9309.

It don't affect functionality. I had take care of it with a small maintainer commit before merging. Overall, the implementation looks good. Thanks again for the contribution!

PR #109, opened 2026-07-18
fix(signals): populate ip_abuse_score correctly and handle list whois dates

01luyicheng:
# fix(signals): populate `ip_abuse_score` correctly and handle list whois dates

Two independent bug fixes in the `full_recon` signals pipeline, plus a small review-feedback commit. Both regressions were introduced by PR #84 (the fullrecon modular refactor) and both actively degrade the threat-analysis output that gets fed to the LLM.

## Summary

| Bug | Effect | Root cause |
| --- | --- | --- |
| `ip_abuse_score` always `0` | Threat-analysis prompt is always fed `IP abuse score : 0/100` regardless of the real AbuseIPDB confidence score — even when the IP is flagged malicious with score 95. | `asn_extractor` looked up `abuse_score` from the `asn_tool` result, but `asn_tool` never returned that field. The canonical score lives in `ip_reputation.abuse_confidence_score`, but `ip_reputation_extractor` never assigned it to the signal. |
| whois `expiration_date` list silently dropped | `domain_expiry_days` stays `None` (rendered as `unknown`) and the `Domain EXPIRED …` / `Domain expires in N days — CRITICAL renewal required` auto-warnings never fire for TLDs where `python-whois` returns a list of dates. | `safe_parse_datetime` did `str(date_input).strip()` on a list, producing `"['2025-12-25 00:00:00']"` which matched no parse format and fell through to `None`. |

Both bugs were found by code review of merged PR #84 (no issue was filed — they are small, well-scoped, and follow the project's "Pull requests are welcome!" policy in `contributing.md`).

## Why This Change

- **Real misinformation to the LLM**: The threat-analysis prompt explicitly reads `IP abuse score : {signals.get('ip_abuse_score', 0)}/100` and `IP flagged malicious: {signals.get('ip_reputation_flagged', False)}`. With the bug, an IP with `abuse_confidence_score=95, is_malicious=True` would be rendered as `IP abuse score : 0/100` and `IP flagged malicious: True` simultaneously — a contradictory, misleading picture. For scores in 1–24, the IP looks completely clean when it isn't.
- **Silent suppression of critical warnings**: The `Domain EXPIRED` / `Domain expires in N days — CRITICAL renewal required` warnings are designed to be high-priority signals for the LLM. When `python-whois` returns a list of expiration dates (which happens for some TLDs/registrars that expose both registry and registrar expiration), the warning is silently dropped.
- **Survived review**: PR #84 went through maintainer + Copilot review and was merged. The signals pipeline had no unit tests at the time (this PR adds the first ones). External follow-up is genuinely valuable here.
- **Small and contained**: ~50 lines of source changes + ~250 lines of tests. No new dependencies, no API changes, fully backward-compatible.
- **No conflict with PRs #106, #107, #108** (different files; this PR touches `tools/signals/*.py`, `utils/helpers.py`, and the corresponding test files).

### Commit 1 — `fix(signals): populate ip_abuse_score from ip_reputation, not asn`

**`tools/signals/ip_reputation.py`** (+4 lines)
- Added `signals["ip_abuse_score"] = rep_score` after the `int(rep_score)` coercion, with an inline comment explaining that `ip_reputation` is the canonical source (AbuseIPDB) and `asn_tool` does not carry this field.

**`tools/signals/asn.py`** (-18 / +9 lines)
- Removed the dead `result.get("abuse_score") or result.get("data", {}).get("abuse_score", 0)` lookup. It always fell through to `0` because `asn_tool` returns `{ip, asn, org, isp, country, region, city}` — no `abuse_score` field.
- Removed the two dead `auto_warnings` branches (`> 50` HIGH and `> 20` elevated). They never fired because `score` was always `0`. The equivalent `MALICIOUS` and `elevated risk` warnings already live in `ip_reputation_extractor`.
- Left a comment explaining why the function is now a no-op for the abuse score (and pointing to `ip_reputation_extractor` as the canonical source).

**`tests/test_signals_extractors.py`** (new file)
- 10 tests covering: ip_reputation_extractor populates score, coerces string, defaults on invalid, skips on failure, MALICIOUS warning, elevated warning; asn_extractor does NOT touch ip_abuse_score, skips on failure; extract_signals integration (score from ip_reputation end-to-end, stays 0 when ip_reputation failed).

### Commit 2 — `fix(helpers): handle list input in safe_parse_datetime for whois dates`

**`utils/helpers.py`** (+16 / -2 lines)
- Added an `isinstance(date_input, list)` branch in `safe_parse_datetime` that:
  - Returns `None` for an empty list.
  - Takes the first element (registry and registrar expiration dates are typically identical; first-seen is the pragmatic choice).
  - Returns `None` if the first element is empty/None.
  - Passthrough if the first element is already a `datetime` object.
- Inline comment explains the `python-whois` TLD variance and why first-element is the right pick.

**`tests/test_helpers.py`** (+49 / -1 lines)
- 8 new tests for `safe_parse_datetime`: ISO string parses; `None` / `""` / `[]` return `None`; `datetime` passthrough; **list of strings takes first element** (regression); list of `datetime` objects takes first; empty list returns `None`; list with empty/None first element returns `None`; unparseable string returns `None`.

### Commit 3 — `review(signals): address cross-review feedback`

Small fixes from multi-agent code review (all MINOR/NIT, non-blocking but worth cleaning up before merge):

**`tests/test_signals_extractors.py`** (+20 / -3 lines)
- Removed unused `whois_extractor` import (the test file covers `asn` and `ip_reputation` extractors; `whois_extractor` is not exercised here).
- Added `test_ip_reputation_extractor_defaults_score_to_zero_when_key_missing` to pin down the defensive-coding contract: when `abuse_confidence_score` is absent from the result dict, `int(None)` raises `TypeError` and the `try/except` falls back to `0`. Previously only the non-numeric-string case was tested.

**`tools/signals/extractor.py`** (+1 / -1 lines)
- Updated the stale comment at line 13 from `# asn` to `# ip_reputation`. The signal is now canonically sourced from `ip_reputation_extractor` (per commit `c1d459c`), so the old `# asn` label was actively misleading.

### What's NOT changed (intentionally)
- **Return shape of `extract_signals`**: Same keys, same initial values. Only the value of `ip_abuse_score` changes from `0` to the real AbuseIPDB score when `ip_reputation` succeeds.
- **Warning semantics**: The `MALICIOUS` and `elevated risk` warnings from `ip_reputation_extractor` are unchanged. The dead `> 50` HIGH warning from `asn_extractor` is removed — it was redundant with the `MALICIOUS` warning (which fires at `abuse_confidence_score >= 25` per `iprep_tool.py`).
- **`asn_extractor` function signature**: Preserved (still callable as `asn_extractor(result, signals)`). The function is now effectively a no-op for the abuse score, with a comment explaining why. Future enhancements (e.g. extracting ASN org/country into signals) can build on it.
- **`safe_parse_datetime` scalar path**: Unchanged. Only list inputs take the new branch. String and datetime inputs follow the exact same code path as before.
- **`whois_extractor` in `tools/signals/whois.py`**: Unchanged — it already calls `safe_parse_datetime(expiry)` correctly; the fix is in the helper, not the caller.

## Verification

```
$ python -m pytest tests/test_signals_extractors.py tests/test_helpers.py tests/test_full_recon.py tests/test_whois_lookup.py -v
39 passed, 10 subtests passed in 0.27s
```

Full test suite (all tools): `193 passed in 0.59s` — no regressions.

Manual reproduction of Bug 1 (before this PR):
```python
# results = {"asn": {"success": True, "ip": "1.2.3.4", ...},
#            "ip_reputation": {"success": True, "is_malicious": True,
#                              "abuse_confidence_score": 95}}
# Before: signals["ip_abuse_score"] == 0  (BUG — even though score is 95)
# After:  signals["ip_abuse_score"] == 95 (FIXED)
```

Manual reproduction of Bug 2 (before this PR):
```python
# safe_parse_datetime(["2025-12-25 00:00:00", "2026-01-01 00:00:00"])
# Before: None  (BUG — str(list) matches no format)
# After:  datetime(2025, 12, 25, 0, 0, 0)  (FIXED — takes first element)
```

## Files Changed

```
 tests/test_helpers.py            |  50 ++++++++-
 tests/test_signals_extractors.py | 220 +++++++++++++++++++++++++++++++++++++++
 tools/signals/asn.py             |  27 ++---
 tools/signals/extractor.py       |   2 +-
 tools/signals/ip_reputation.py   |   4 +
 utils/helpers.py                 |  18 +++-
 6 files changed, 299 insertions(+), 22 deletions(-)
```

## Checklist

- [x] Both fixes follow the existing `try/except` + `{"success": ...}` pattern from `contributing.md` (the extractors and helpers already follow it).
- [x] No new dependencies; no API changes; backward-compatible.
- [x] All pre-existing tests in `test_helpers.py`, `test_full_recon.py`, `test_whois_lookup.py` still pass; full suite (193 tests) passes.
- [x] 11 new tests for signals extractors + 8 new tests for `safe_parse_datetime` = 19 new tests.
- [x] No real network calls in tests (all extractors tested with plain dicts).
- [x] `signals` dict shape preserved — only the value of `ip_abuse_score` changes from `0` to the real AbuseIPDB score when `ip_reputation` succeeds.
- [x] `safe_parse_datetime` scalar path unchanged — only list inputs take the new branch.
- [x] Inline comments explain the canonical-source decision and the `python-whois` TLD variance.
- [x] Minimal diff — no scope creep, no bundled unrelated cleanup. Commit 3 is strictly review feedback.

## Related

- Follow-up to PR #84 by @gaoharimran29-glitch (the fullrecon modular refactor that introduced both regressions).
- No issue was filed — both bugs are small and well-scoped enough to fix directly per `contributing.md`'s "Pull requests are welcome!" policy.
- No conflict with PRs #106, #107, #108 (different files).

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -1,4 +1,5 @@
-from utils.helpers import is_valid_domain , get_cvss_details , get_english_description
+from utils.helpers import is_valid_domain , get_cvss_details , get_english_description , safe_parse_datetime
+from datetime import datetime, timezone
 import unittest
 
 class TestHelpers(unittest.TestCase):
@@ -54,6 +55,53 @@ class TestHelpers(unittest.TestCase):
 
     def test_empty_descriptions_returns_empty(self):
         self.assertEqual(get_english_description({}), "")
+
+    # ── safe_parse_datetime ─────────────────────────────────
+
+    def test_safe_parse_datetime_iso_string(self):
+        """A standard ISO datetime string parses correctly."""
+        result = safe_parse_datetime("2025-12-25 00:00:00")
+        self.assertEqual(result, datetime(2025, 12, 25, 0, 0, 0))
+
+    def test_safe_parse_datetime_none_input(self):
+        """None or empty input returns None."""
+        self.assertIsNone(safe_parse_datetime(None))
+        self.assertIsNone(safe_parse_datetime(""))
+        self.assertIsNone(safe_parse_datetime([]))
+
+    def test_safe_parse_datetime_passthrough_datetime_object(self):
+        """A datetime object is returned as-is."""
+        dt = datetime(2025, 12, 25, 0, 0, 0, tzinfo=timezone.utc)
+        self.assertEqual(safe_parse_datetime(dt), dt)
+
+    def test_safe_parse_datetime_list_takes_first_element(self):
+        """A list of date strings (python-whois TLD variance) takes the first.
+
+        Regression test for the bug where str(list) produced
+        "['2025-12-25 00:00:00']" which matched no format, silently
+        suppressing domain-expiry warnings in whois_extractor.
+        """
+        result = safe_parse_datetime(["2025-12-25 00:00:00", "2026-01-01 00:00:00"])
+        self.assertEqual(result, datetime(2025, 12, 25, 0, 0, 0))
+
+    def test_safe_parse_datetime_list_of_datetime_objects(self):
+        """A list of datetime objects takes the first element as-is."""
+        dt1 = datetime(2025, 12, 25, 0, 0, 0, tzinfo=timezone.utc)
+        dt2 = datetime(2026, 1, 1, 0, 0, 0, tzinfo=timezone.utc)
+        self.assertEqual(safe_parse_datetime([dt1, dt2]), dt1)
+
+    def test_safe_parse_datetime_empty_list_returns_none(self):
+        """An empty list returns None, not a crash."""
+        self.assertIsNone(safe_parse_datetime([]))
+
+    def test_safe_parse_datetime_list_with_empty_first_element(self):
+        """A list whose first element is empty/None falls through to None."""
+        self.assertIsNone(safe_parse_datetime([None, "2025-12-25"]))
+        self.assertIsNone(safe_parse_datetime(["", "2025-12-25"]))
+
+    def test_safe_parse_datetime_invalid_string_returns_none(self):
+        """A genuinely unparseable string returns None, not a crash."""
+        self.assertIsNone(safe_parse_datetime("not-a-date"))
 
 if __name__ == "__main__":
     unittest.main(verbosity=2)
--- a/tests/test_signals_extractors.py
+++ b/tests/test_signals_extractors.py
@@ -1,0 +1,204 @@
+"""Tests for the per-tool signal extractors in tools/signals/.
+
+These tests pin down the contract that each extractor mutates the shared
+`signals` dict in the way the threat-analysis prompt expects. They are
+especially important for the ip_abuse_score signal, which was silently
+always 0 between PR #84 and the fix that moved the assignment into
+ip_reputation_extractor (the AbuseIPDB-backed canonical source).
+"""
+from tools.signals.asn import asn_extractor
+from tools.signals.ip_reputation import ip_reputation_extractor
+from tools.signals.whois import whois_extractor
+from tools.signals.extractor import extract_signals
+
+
+def _base_signals():
+    """Return a fresh signals dict matching extract_signals' initial shape."""
+    return {
+        "domain_expiry_days": None,
+        "dns_missing_records": [],
+        "open_ports": [],
+        "ssl_days_remaining": None,
+        "software_detected": [],
+        "ip_abuse_score": 0,
+        "subdomain_count": 0,
+        "missing_security_headers": [],
+        "email_security": {},
+        "ip_reputation_flagged": False,
+        "auto_warnings": [],
+    }
+
+
+# ── ip_reputation_extractor ──────────────────────────────────────────────
+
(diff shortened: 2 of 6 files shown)


gaoharimran29-glitch (approved):
@01luyicheng Thanks for the thorough contribution! I verified both issues and the fixes look good. The ip_abuse_score bug was straightforward to reproduce, and after digging through the WHOIS signal pipeline, the list-handling issue in safe_parse_datetime() is indeed a legitimate edge case that suppresses domain expiry warnings for certain WHOIS responses.

I also went through the added regression tests and appreciate the level of coverage. We didn't previously have dedicated extractor tests in the repository, but the tests added here are well-scoped and improve coverage for the signals pipeline, so I'm happy to adopt them going forward.

Thanks for taking the time to document the root causes, provide manual reproductions, and keep the changes focused and backward-compatible. This is a well-prepared bug-fix PR. I'll be merging it.

PR #112, opened 2026-07-19
feat(robots): preserve per-user-agent rules and crawl-delay groups

### LINKED ISSUE #110: Enhancement: `robots_txt_inspect` to preserve per-User-agent directives and rule groups
## Context

The `robots_txt_inspect` tool currently parses `User-agent`, `Allow`, `Disallow`, `Sitemap`, `Crawl-delay`, and `Host` directives. While `Crawl-delay` is now parsed correctly, the parser does not preserve per-User-agent directives or rule groups that do not contain `Allow` or `Disallow` entries.

For example, the following `robots.txt` file is perfectly valid and commonly used by websites:

```txt
User-agent: SemrushBot*
Crawl-delay: 10

User-agent: Applebot
Crawl-delay: 10

User-agent: Bingbot
Crawl-delay: 10

User-agent: AhrefsBot
Crawl-delay: 10
```

The tool currently returns only a top-level value:

```json
{
    "crawl_delay": "10"
}
```

All individual `User-agent` blocks are lost because the parser only persists rule groups containing `Allow` or `Disallow` directives.

---

## Why This Matters

`Crawl-delay` is interpreted on a per-User-agent basis. Many websites specify different crawling instructions for different bots without defining any `Allow` or `Disallow` rules.

As a result, valuable metadata is currently discarded.

The robots.txt specification also allows multiple consecutive `User-agent` directives to belong to the same rule group. For example:

```txt
User-agent: Googlebot
User-agent: Googlebot-Image

Disallow: /private/
```

Both User-agents should belong to the same rule group and inherit the same directives.

The current implementation does not preserve this behavior.

---

## Proposed Improvements

- Preserve all User-agent rule groups, even if they only contain directives such as `Crawl-delay`.
- Store `crawl_delay` inside each rule block
- Support multiple consecutive `User-agent` directives belonging to the same rule group.
- Continue exposing aggregated top-level fields such as:
  - `allowed_paths`
  - `disallowed_paths`
  - `sitemaps`
- Maintain backward compatibility wherever possible.

---

## Proposed Output

Instead of:

```json
{
    "crawl_delay": "10"
}
```

Return:

```json
{
    "rules": [
        {
            "user_agent": "SemrushBot*",
            "allow": [],
            "disallow": [],
            "crawl_delay": "10"
        },
        {
            "user_agent": "Applebot",
            "allow": [],
            "disallow": [],
            "crawl_delay": "10"
        },
        {
            "user_agent": "Bingbot",
            "allow": [],
            "disallow": [],
            "crawl_delay": "10"
        },
        {
            "user_agent": "AhrefsBot",
            "allow": [],
            "disallow": [],
            "crawl_delay": "10"
        }
    ]
}
```

For grouped User-agents:

```txt
User-agent: Googlebot
User-agent: Googlebot-Image

Disallow: /private/
```

The output could be represented as:

```json
{
    "rules": [
        {
            "user_agents": [
                "Googlebot",
                "Googlebot-Image"
            ],
            "allow": [],
            "disallow": [
                "/private/"
            ],
            "crawl_delay": null
        }
    ]
}
```

---

## Notes

- This is an enhancement request and not a regression introduced by the recent `Crawl-delay` parsing changes.
- Websites such as `inomics.com` provide good real-world examples for testing this behavior.
- The implementation should preserve crawler-specific directives and accurately represent robots.txt rule groups.

Solaris-star:
## Summary
Closes #110

`robots_txt_inspect` previously only kept rule groups that contained Allow/Disallow, so common crawl-delay-only User-agent blocks were dropped.

### Changes
- Preserve all User-agent groups, including crawl-delay-only blocks
- Support consecutive `User-agent` lines sharing the same directives
- Store `crawl_delay` on each rule
- Keep top-level `allowed_paths` / `disallowed_paths` / `sitemaps` / last-seen `crawl_delay` for compatibility

### Example
```txt
User-agent: SemrushBot*
Crawl-delay: 10

User-agent: Applebot
Crawl-delay: 10
```
now returns per-agent rules instead of only a single top-level crawl_delay.

## Testing
```bash
uv run pytest tests/test_robots_txt_tool.py -v
# 9 passed
```

### DIFF
--- a/tools/robots_txt_tool.py
+++ b/tools/robots_txt_tool.py
@@ -1,6 +1,7 @@
 import requests
 from utils.helpers import is_valid_domain
 
+
 def robots_txt_inspect(domain: str) -> dict:
     """
     Fetch and parse the robots.txt file for a given domain to reveal hidden directories and sitemaps.
@@ -12,7 +13,7 @@ def robots_txt_inspect(domain: str) -> dict:
         headers = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"}
         url_https = f"https://{domain}/robots.txt"
         url_http = f"http://{domain}/robots.txt"
-        
+
         response = None
         try:
             response = requests.get(url_https, timeout=10.0, headers=headers)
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch:
@SemTiOne Hi! I've assigned you as a reviewer on a couple of PRs. We've had quite a few contributions recently, and I'd appreciate your help with reviewing them when you have some time.

tools/robots_txt_tool.py
```diff
@@ -1,6 +1,7 @@
 import requests
 from utils.helpers import is_valid_domain
```
SemTiOne, line 2: `normalize_domain` was dropped from this import but it's not mentioned anywhere in the PR description. This is a regression, not cleanup (see comment on line 9).

tools/robots_txt_tool.py
SemTiOne, line 10: This used to have `domain = normalize_domain(domain)` right before the validation check. It's gone now. Effect:
```python
robots_txt_inspect('example.com.')   # trailing dot, valid FQDN form
# -> {'success': False, 'error': 'Invalid domain format'}

robots_txt_inspect(' example.com ')  # whitespace
# -> {'success': False, 'error': 'Invalid domain format'}
```
Both work fine on `main` before this PR. Every other tool in this repo (`dns_tool.py`, `headers_tool.py`, `ssl_tool.py`, `cloud_exposure_tool.py`, `crt_sh_tool.py`, `email_security_tool.py`, `fullrecon_tool.py`, `redirect_tracer.py`, `techstack_tool.py`, `whois_tool.py`) still calls `normalize_domain()`, so this tool is now the only inconsistent one. Since it's registered directly on the MCP server with no wrapper (`mcp.tool()(robots_txt_inspect)`), this silently breaks any client sending a domain pulled from free text.

Please restore:
```python
if not is_valid_domain(domain):
```
plus a test case covering trailing dot and whitespace so this doesn't regress again.

tools/robots_txt_tool.py
```diff
@@ -21,90 +22,106 @@ def robots_txt_inspect(domain: str) -> dict:
...
+                        "disallow": list(disallow),
+                        "crawl_delay": current_crawl_delay,
+                    }
+                )
+            current_agents = []
+            current_allow = []
+            current_disallow = []
+            current_crawl_delay = None
+            seen_directive_in_group = False
+
```
SemTiOne, line 63: If a directive (`Disallow:`/`Allow:`/`Crawl-delay:`) appears before the first `User-agent` line in a malformed robots.txt, `current_agents` is still empty so `flush_group()` returns early at `if not current_agents: return`, without resetting `current_allow`/`current_disallow`. That orphaned directive then gets silently attached to the first real User-agent group that follows:
```python
# robots.txt:
# Disallow: /orphan-should-not-count
# User-agent: A
# Disallow: /a-real-path

result['rules'][0]
# {'user_agent': 'A', 'disallow': ['/orphan-should-not-count', '/a-real-path'], ...}
```

The old code was also wrong here (it defaulted the orphan to "*"), but at least it didn't misattribute it to a specific named bot.

SemTiOne (changes requested):
Please fix all the requested changes @Solaris-star.

Solaris-star:

PR #114, opened 2026-07-19
feat(ip_reputation): richer classification and AbuseIPDB metadata

### LINKED ISSUE #101: Enhancement: Improve IP Reputation Classification by Including AbuseIPDB Whitelist and Reputation Metadata
# Summary

The current `ip_reputation` tool determines whether an IP address is malicious solely based on the `abuseConfidenceScore` returned by the AbuseIPDB API (`abuseConfidenceScore >= 25`).

This approach can lead to misleading results because AbuseIPDB provides additional reputation-related metadata that is currently ignored. For example, Google's public DNS server (`8.8.8.8`) returns:

* `abuseConfidenceScore: 0`
* `isWhitelisted: True`
* `totalReports: 120`
* `numDistinctUsers: 65`
* `usageType: Content Delivery Network`

Although the IP has been reported multiple times, it is explicitly whitelisted by AbuseIPDB and should be treated as a trusted service rather than simply being marked as "not malicious."

Also currently the tool dont have load_dotenv() to load the .env file

### Proposed Changes

* Include additional fields from the AbuseIPDB API response:

  * `isWhitelisted`
  * `isPublic`
  * `isTor`
  * `numDistinctUsers`
  * `hostnames`
  * `ipVersion`
 

* Introduce a human-readable reputation classification instead of relying only on a boolean flag. For example:

  * `trusted` (whitelisted IPs)
  * `high-risk` (high abuse confidence score)
  * `suspicious`
  * `low-risk`
  * `clean`

* Preserve the existing `abuse_confidence_score` and report counts for transparency.
* Add api_key loading use load-dotenv()

### Why This Change?

Using only `abuseConfidenceScore` hides valuable context provided by AbuseIPDB and may produce incomplete or misleading results for well-known services, CDNs, and public infrastructure IPs. Exposing the additional metadata and providing a more descriptive reputation classification will make the tool significantly more useful for security and reconnaissance workflows.

Files to touch :- tools/iprep_tool.py
tests/test_ip_reputation.py

Solaris-star:
## Summary
Closes #101

`ip_reputation` previously reduced AbuseIPDB results to a single boolean based on `abuseConfidenceScore >= 25`. That hides useful context (for example Google DNS can have reports but be whitelisted).

### Changes
- Add human-readable `reputation`:
  - `trusted` (whitelisted)
  - `high-risk` (score >= 75)
  - `suspicious` (score >= 25)
  - `low-risk` (reports with low/zero score)
  - `clean`
- Include additional AbuseIPDB fields:
  - `is_whitelisted`, `is_public`, `is_tor`
  - `num_distinct_users`, `hostnames`, `ip_version`
  - keep existing score/report/country/isp/domain/usage_type/last_reported_at
- Preserve legacy `is_malicious` (whitelist overrides malicious)

## Testing
```bash
uv run pytest tests/test_ip_reputation.py -v
# 9 passed
```

### DIFF
--- a/tests/test_ip_reputation.py
+++ b/tests/test_ip_reputation.py
@@ -1,10 +1,11 @@
-from unittest.mock import Mock, MagicMock, patch, call
-import unittest
-from tools.iprep_tool import ip_reputation
 import os
+import unittest
+from unittest.mock import Mock, patch
+
+from tools.iprep_tool import _classify_reputation, ip_reputation
+
 
 class TestIpReputation(unittest.TestCase):
-
     def test_ip_reputation_requires_valid_ip_and_api_key(self):
         self.assertFalse(ip_reputation("not-an-ip")["success"])
 
@@ -23,6 +24,12 @@ class TestIpReputation(unittest.TestCase):
                 "ipAddress": "1.2.3.4",
                 "abuseConfidenceScore": 95,
                 "totalReports": 342,
+                "numDistinctUsers": 40,
+                "isWhitelisted": False,
+                "isPublic": True,
+                "isTor": False,
+                "ipVersion": 4,
+                "hostnames": ["bad.example"],
                 "countryCode": "CN",
                 "isp": "Example ISP",
                 "domain": "example.net",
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (changes requested):
@Solaris-star Thanks for this the whitelist-first approach and the added metadata fields are exactly what we needed. A small suggestion things before merge:

## 1. Report-volume signal gets lost when `abuse_score` decays to 0

```python
if total_reports > 0 and abuse_score > 0:
    return "low-risk"
if total_reports > 0:
    return "low-risk"
```

Both branches return the same thing, so an IP with `total_reports = 500, abuse_score = 0` gets classified identically to one with `total_reports = 1, abuse_score = 0`. AbuseIPDB scores decay over time even when report volume is high, so this collapses a meaningful distinction — high historical report volume is still useful signal for a security workflow, even at score 0.

Can you either:
- add a report-count threshold that bumps to `suspicious` (or a new label) when reports are high but score is 0, **or**
- confirm this flattening is intentional per the original spec, and if so just collapse the duplicate branch into one (`if total_reports > 0: return "low-risk"`) so it's not dead code.

Otherwise looks good — the classification structure and metadata exposure are a solid improvement over the boolean-only approach.

Solaris-star:

PR #126, opened 2026-07-22
ci: remove unused labeler.yml, expand issue label keyword matching

### LINKED ISSUE #77: CI/CD: Remove unused .github/labeler.yaml and expand automatic issue labeling in .github/workflows.yaml
## Summary

The repository currently contains two label-related files:

- `.github/labeler.yaml`
- `.github/workflows/labeler.yaml`

However, the workflow in `.github/workflows/labeler.yaml` contains all label matching logic directly in the `github-script` step and does not read or reference `.github/labeler.yaml`.

As a result, `.github/labeler.yaml` is unused and can be misleading for contributors, who may assume it is part of the labeling workflow.

---

## Why this matters

Maintaining an unused configuration file increases repository complexity and can lead to confusion when updating or extending the automatic labeling logic.

Additionally, the current workflow only recognizes a small set of keywords and labels (`bug`, `enhancement`, `documentation`, `good first issue`, and `question`). Issues that should receive other common labels (for example, `security`, `performance`, `dependencies`, `testing`, `ci`, `docker`, `refactor`, `reliability`, etc.) are not automatically labeled.

---

## Suggested fix

- Remove the unused `.github/labeler.yaml` file.
- Expand the keyword matching logic in `.github/workflows/labeler.yaml` to support all issue labels used by the repository.
- Add appropriate keyword aliases for each label to improve automatic issue categorization.
- Ensure the workflow remains easy to maintain as new labels are introduced.

---

## Acceptance criteria

- [ ] Remove the unused `.github/labeler.yaml` file
- [ ] Extend the issue labeling workflow to cover all repository issue labels
- [ ] Add sensible keyword aliases for each supported label
- [ ] Verify issues are automatically labeled correctly based on their titles

---

## Files affected

- `.github/labeler.yaml`
- `.github/workflows/labeler.yaml`

SemTiOne:
Fixes #77

## What changed
- Removed `.github/labeler.yml` — dead config, no workflow reads it; matching logic already lives inline in the workflow file.
- Refactored `.github/workflows/labeler.yml` from chained if-statements to a single `LABEL_KEYWORDS` map, and expanded coverage from 5 to 7 labels: `bug`, `enhancement`, `good first issue`, `documentation`, `question`, `cleanup`, `reliability`.

## Why only 7, not all 18 repo labels
Skipped on purpose:
- `dependencies`, `dockerfile`, `python:uv` — Dependabot-owned, PR-scoped labels. This workflow only fires on `issues` events, so they'd never apply here anyway.
- `duplicate`, `invalid`, `wontfix`, `help wanted` — maintainer triage calls, not something a title keyword can safely determine.
- `size: small/medium/large` — already handled by `pr-size-labeler.yml`.
- `Stale` — handled by a separate time-based workflow.

## Testing
Ran the extracted logic locally against sample issue titles, including this issue's own title, to confirm correct label assignment. No CI test harness exists for this workflow, so verification was manual. Happy to add a test fixture if you'd like one.


gaoharimran29-glitch (approved):
@SemTiOne Thanks for the contribution!

The refactor makes the workflow much easier to maintain and extend, and the rationale for excluded labels is clearly documented. The keyword map approach is significantly cleaner than maintaining multiple conditional blocks.

Looks good to me. Thanks for the improvement!

PR #132, opened 2026-07-25
feat(ip-reputation): add reputation classification and expose AbuseIP…

### LINKED ISSUE #101: Enhancement: Improve IP Reputation Classification by Including AbuseIPDB Whitelist and Reputation Metadata

gaoharimran29-glitch:
## Summary

This PR enhances the `ip_reputation` tool by introducing a human-readable reputation classification model, exposing additional AbuseIPDB metadata, and improving environment variable loading through `.env` support.

Previously, the tool determined whether an IP address was malicious solely based on the `abuseConfidenceScore` threshold (`>= 25`). While functional, this approach ignored important context provided by AbuseIPDB such as whitelisting status, Tor usage, and public infrastructure indicators.

Additionally, the tool relied entirely on environment variables being manually exported, which could lead to configuration issues during local development.

This change provides richer reputation analysis while preserving backward compatibility through the existing `is_malicious` field and improves developer experience by supporting `.env`-based configuration.

---

### Reputation Classification

Added a dedicated reputation classification model that categorizes IPs into:

* [x] `trusted`
* [x] `high-risk`
* [x] `suspicious`
* [x] `low-risk`
* [x] `clean`

Classification logic now considers:

* Abuse confidence score
* AbuseIPDB whitelist status
* Tor exit node status

---

### Additional AbuseIPDB Metadata

Exposed the following fields in the tool output:

* [x] `ip_version`
* [x] `is_whitelisted`
* [x] `is_public`
* [x] `is_tor`
* [x] `num_distinct_users`
* [x] `hostnames`
* [x] `reputation`

Existing fields such as:

* [x] `abuse_confidence_score`
* [x] `total_reports`

remain available for transparency and compatibility.

---

### Environment Configuration Improvements

Added support for loading environment variables from a `.env` file using `python-dotenv`.

Changes include:

* [x] Added `python-dotenv` dependency
* [x] Added automatic `.env` loading during tool initialization
* [x] Improved local development experience by reducing manual environment setup requirements
* [x] Maintained existing support for standard environment variables

This ensures the AbuseIPDB API key can be loaded consistently across local development and deployment environments.

---

### Backward Compatibility

The existing `is_malicious` field has been retained.

Instead of relying directly on the abuse confidence score threshold, it is now derived from the reputation classification:

* `high-risk` → malicious
* `suspicious` → malicious
* `trusted` → not malicious
* `low-risk` → not malicious
* `clean` → not malicious

This ensures existing integrations continue to work without modification.

---

File Changes:-
* [x] Tests updated
* [x] Dependecies file updated
* [x] Tool file updated

Closes #101

### DIFF
--- a/tools/iprep_tool.py
+++ b/tools/iprep_tool.py
@@ -1,6 +1,29 @@
 import ipaddress
 import requests
 import os
+from dotenv import load_dotenv
+
+load_dotenv()
+
+def classify_reputation(is_whitelisted: bool, abuse_score: int, is_tor: bool) -> str:
+    """Classification model to classify the reputation of the 
+    ip on the basis on is_whitelisted, abuse_score and is_tor"""
+    if is_whitelisted:
+        return "trusted"
+
+    if abuse_score >= 75:
+        return "high-risk"
+
+    if is_tor:
+        return "suspicious"
+
+    if abuse_score >= 25:
+        return "suspicious"
+
+    if abuse_score > 0:
+        return "low-risk"
+
+    return "clean"
 
 def ip_reputation(ip_address: str) -> dict:
     """
@@ -30,15 +53,26 @@ def ip_reputation(ip_address: str) -> dict:
         data = response.json().get("data", {})
 
         abuse_score = data.get("abuseConfidenceScore", 0)
+        is_whitelisted = bool(data.get("isWhitelisted" , False))
+        is_tor = bool(data.get("isTor" , False))
+        reputation = classify_reputation(is_whitelisted, abuse_score, is_tor)
+
         return {
             "success": True,
             "ip": ip,
-            "is_malicious": abuse_score >= 25,
+            "ip_version": data.get("ipVersion"),
+            "is_malicious": reputation in {"high-risk" , "suspicious"},
+            "is_whitelisted": is_whitelisted,
+            "is_public": bool(data.get("isPublic" , False)),
+            "is_tor": is_tor,
+            "reputation": reputation,
+            "num_distinct_users": data.get("numDistinctUsers" , 0),
             "abuse_confidence_score": abuse_score,
             "total_reports": data.get("totalReports", 0),
             "country": data.get("countryCode"),
             "isp": data.get("isp"),
             "domain": data.get("domain"),
+            "hostnames": data.get("hostnames"),
             "usage_type": data.get("usageType"),
             "last_reported_at": data.get("lastReportedAt"),
         }
--- a/tools/iprep_tool.py
+++ b/tools/iprep_tool.py
@@ -66,7 +66,7 @@ def ip_reputation(ip_address: str) -> dict:
             "is_public": bool(data.get("isPublic" , False)),
             "is_tor": is_tor,
             "reputation": reputation,
-            "num_distinct_users": data.get("numDistinctUsers" , 0),
+            "num_distinct_users": data.get("numDistinctUsers") or 0,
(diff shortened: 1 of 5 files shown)


tools/iprep_tool.py
```diff
@@ -30,15 +53,26 @@ def ip_reputation(ip_address: str) -> dict:
...
             "success": True,
             "ip": ip,
-            "is_malicious": abuse_score >= 25,
+            "ip_version": data.get("ipVersion"),
+            "is_malicious": reputation in {"high-risk" , "suspicious"},
+            "is_whitelisted": is_whitelisted,
+            "is_public": bool(data.get("isPublic" , False)),
+            "is_tor": is_tor,
+            "reputation": reputation,
+            "num_distinct_users": data.get("numDistinctUsers" , 0),
```
SemTiOne: Not None-safe, unlike the `bool(...)` wraps on lines 56/57/66 for is_whitelisted/is_tor/is_public. `.get(key, default)` doesn't catch explicit `None` values, only missing keys. `test_handles_null_fields_gracefully` proves AbuseIPDB returns explicit `None` for these fields. This returns `None` instead of `0` when `numDistinctUsers` is `None`, will `TypeError` on any numeric comparison downstream. Fix: `data.get("numDistinctUsers") or 0`.

SemTiOne (changes requested):
Solid work. One blocking issue on num_distinct_users (see inline comment).

tools/iprep_tool.py
gaoharimran29-glitch: Good catch, thanks. Updated the handling to default num_distinct_users to 0 when AbuseIPDB returns null and pushed the change in commit `ef8e895`

SemTiOne (approved):
LGTM!

PR #133, opened 2026-07-26
feat: add Imperva/Incapsula body-signature detection + Cloudflare 403 edge-case test

SemTiOne:
Related to #62 (continued from #68, #131)

## Problem
Imperva/Incapsula block pages were undetected, only 1 sample existed, too small to act on.

## Fix
Probed 36 domains, found 1 new Imperva hit (geico.com). Compared browser-impersonated vs plain request to trigger its actual block page (no attack payload). Found X-Iinfo/X-CDN headers present on both normal and blocked responses, can't use headers alone like Cloudflare/Akamai. The block page's body embeds `_Incapsula_Resource`.

Extended detection to check body content, not just headers. `_walk_redirect_chain` now captures `resp.text` (already downloaded, no extra cost). Split signatures into header-based and body-based tables. Imperva requires `status_code >= 400`, same defensive pattern as Akamai.

## Validation
```bash
pytest tests/test_headers_tool.py -v --cov=tools.headers_tool
51 passed, 100% coverage
```
4 new tests: 3 for Imperva (block-page test confirmed failing pre-fix, passing post-fix), 1 confirming a Cloudflare origin's real 403 without `cf-mitigated` still analyzes normally.

## Does NOT resolve
- Sucuri: 0 samples, wrong domain category.
- AWS/CloudFront, DataDome: no confirmed block page yet.
- Imperva: only 1 live sample (third-party corroboration, not a second first-hand confirmation).

### DIFF
--- a/tests/test_headers_tool.py
+++ b/tests/test_headers_tool.py
@@ -5,7 +5,7 @@ from curl_cffi.requests.headers import Headers
 from tools.headers_tool import headers_analyzer
 
 
-def _resp(status_code: int, headers: dict):
+def _resp(status_code: int, headers: dict, body: str = ""):
     """Build a mock curl_cffi response for a single hop.
 
     Wraps headers in curl_cffi's real Headers class (not a plain dict)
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
Thanks for the follow-up work here.

I like that the Imperva detection is based on investigation rather than assumptions. Requiring both the `_Incapsula_Resource` marker and a >=400 status code keeps the heuristic conservative and reduces false-positive risk. The separation between header-based and body-based signatures also keeps the detection logic easy to extend,. LGTM.

PR #134, opened 2026-07-26
fix(crtsh): remove passive DNS fallback

### LINKED ISSUE #104: Refactor: `cert_transparency` Tool to Perform Pure Certificate Transparency Lookups

gaoharimran29-glitch:
## Summary

Fix the `cert_transparency` tool to perform only Certificate Transparency lookups via `crt.sh`.

Previously, the tool fell back to HackerTarget's passive DNS API when `crt.sh` failed. Since HackerTarget does not provide Certificate Transparency data, this could return passive DNS results while exposing CT-specific fields such as certificate issuers, validity dates, and certificate counts, leading to misleading results.

Closes #104

## Changes

- Removed the HackerTarget fallback implementation.
- Restricted the tool to Certificate Transparency lookups through `crt.sh`.
- Removed the unused `impersonate="chrome"` request argument.
- Added explicit error handling for:
  - Network/request failures
  - Invalid JSON responses
  - Generic Exceptions
- Return clear structured errors when CT data cannot be retrieved.
- Updated `mcp.json` tool metadata.
- Updated README documentation to reflect the new behavior.
- Added/updated tests covering successful and failure scenarios.

### Updates in test file 

The main changes were:
- Removed the old fallback-related tests that expected a second request path and a source value of hackertarget_fallback. That path no longer exists in the tool, so those tests were failing for the wrong reason.
- Expanded the success test to verify the actual crt.sh response shape now returned by the tool: source, domain, unique_subdomains, wildcards_found, total_unique_subdomains, total_certificates, returned_certificates, truncated, and certificates.
- Tightened the invalid-domain test so it checks the exact error message, not just success being false.
- Replaced the removed fallback tests with direct request-error coverage, which matches the current exception handling in the tool.
- Kept and adjusted the wildcard tests so they assert the current wildcard normalization and filtering behavior.
- Added a truncation test to confirm the 50-certificate limit is enforced and reported correctly.

## Testing

- Verified successful Certificate Transparency lookups return:
  - Certificate issuer information
  - Validity dates
  - Wildcard certificates
  - Unique subdomains
  - Certificate counts
- Verified invalid domains return validation errors.
- Ran test suite and confirmed all tests pass.

## Checklist

- [x] Removed non-CT fallback behavior
- [x] Preserved existing CT response schema
- [x] Added broad exception handling 
- [x] Added/updated tests
- [x] Updated documentation
- [x] Updated MCP metadata

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ It is also listed on glama mcp registry.
 | `ssl_inspect` | SSL/TLS certificate — issuer, expiry, cipher strength, SANs, TLS version |
 | `email_security_check` | Checks SPF, DKIM, and DMARC DNS records — returns a security_score, rating, and actionable recommendations for missing or weak configurations |
 | `tech_stack_detect` | Web server, CMS, JS frameworks, CDN, analytics, and security header scoring |
-| `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information" |
+| `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information |
 | `asn_lookup` | Autonomous System Number (ASN) and network ownership lookup via Team Cymru WHOIS — identifies ASN, BGP prefix, organization, registry, country, and allocation date for domains or IP addresses (no API key required) |
 | `ip_reputation` | Check if an IP is flagged as malicious via AbuseIPDB (api key requied) |
 | `full_recon` | Runs all core tools in parallel and returns combined result |
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ It is also listed on glama mcp registry.
 | `ssl_inspect` | SSL/TLS certificate — issuer, expiry, cipher strength, SANs, TLS version |
 | `email_security_check` | Checks SPF, DKIM, and DMARC DNS records — returns a security_score, rating, and actionable recommendations for missing or weak configurations |
 | `tech_stack_detect` | Web server, CMS, JS frameworks, CDN, analytics, and security header scoring |
-| `cert_transparency` | Subdomain discovery via crt.sh Certificate Transparency logs with an automatic fallback to HackerTarget passive DNS on timeouts |
+| `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information" |
 | `asn_lookup` | Autonomous System Number (ASN) and network ownership lookup via Team Cymru WHOIS — identifies ASN, BGP prefix, organization, registry, country, and allocation date for domains or IP addresses (no API key required) |
 | `ip_reputation` | Check if an IP is flagged as malicious via AbuseIPDB (api key requied) |
 | `full_recon` | Runs all core tools in parallel and returns combined result |
--- a/tools/crt_sh_tool.py
+++ b/tools/crt_sh_tool.py
@@ -1,40 +1,17 @@
 from curl_cffi import requests
-from curl_cffi.requests.errors import RequestsError 
-from typing import Dict, Any, Set
-import requests as standard_requests  # Used for the quick fallback API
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -47,7 +47,7 @@ It is also listed on glama mcp registry.
 | `ssl_inspect` | SSL/TLS certificate — issuer, expiry, cipher strength, SANs, TLS version |
 | `email_security_check` | Checks SPF, DKIM, and DMARC DNS records — returns a security_score, rating, and actionable recommendations for missing or weak configurations |
 | `tech_stack_detect` | Web server, CMS, JS frameworks, CDN, analytics, and security header scoring |
-| `cert_transparency` | Subdomain discovery via crt.sh Certificate Transparency logs with an automatic fallback to HackerTarget passive DNS on timeouts |
+| `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information" |
```
SemTiOne:
```suggestion
| `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information |
```

tools/crt_sh_tool.py
```diff
@@ -47,8 +24,7 @@ def cert_transparency(domain: str) -> Dict[str, Any]:
     headers = {"User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36"}
 
     try:
-        # Try primary source: crt.sh with Chrome impersonation
-        response = requests.get(url, headers=headers, timeout=50, impersonate="chrome")
+        response = requests.get(url, headers=headers, timeout=50)
```
SemTiOne: Calling `impersonate="chrome"` unused isn't quite right, it was on every request. Worth confirming `crt.sh` still behaves the same without it in practice, not just in the mocked tests.

SemTiOne (approved):
Well done! Two small notes inline, non-blocking.

tools/crt_sh_tool.py
gaoharimran29-glitch: I am re adding `impersonate="chrome"` . This preserves browser-like request fingerprints and helps avoid any issues with sites that perform TLS/browser fingerprint checks. Thanks for catching out.

gaoharimran29-glitch:
@SemTiOne Thanks for review . i addressed your both concerns and going to merge this.

PR #137, opened 2026-07-31
feat: support grouped user-agents and per-rule crawl-delay

### LINKED ISSUE #110: Enhancement: `robots_txt_inspect` to preserve per-User-agent directives and rule groups

gaoharimran29-glitch:
**Summary**

This PR updates the robots.txt parser in robots_txt_tool.py to produce cleaner rule groups and to store directive metadata at the rule level instead of as a single top-level `crawl_delay` value. It also refreshes test_robots_txt_tool.py so the test suite matches the new parsing behavior and covers the new grouping logic.

**What changed**

- Added a small `save_rule()` helper in robots_txt_tool.py to normalize and append parsed rule blocks in one place.
- Updated the parser so consecutive `User-agent` lines can belong to the same rule group until a directive is seen.
- Moved `crawl_delay` into each rule object instead of returning it as a top-level field.
- Kept support for `allow`, `disallow`, `sitemaps`, and `host`, while preserving top-level aggregated `allowed_paths` and `disallowed_paths` for compatibility.
- Added a short docstring to `save_rule()` to clarify its responsibility.
- Updated tests in test_robots_txt_tool.py to assert the new per-rule `crawl_delay` shape.
- Added a regression test for grouped multi-agent rules to confirm that multiple `User-agent` directives are captured in one rule block.

**Breaking Changes**

- `crawl_delay` moved from top-level to per-User-agent-group.

Previously: `result["crawl_delay"]`
Now: `result["rules"][i]["crawl_delay"]`

Rationale: A single robots.txt can specify different crawl delays for different user agents, so a single top-level value was `inaccurate/lossy` when multiple groups had different delays. 

Impact: Any consumer reading `result["crawl_delay"]` directly will now get a `KeyError` instead of a value. 

Migration: Consumers should read the crawl delay from the relevant group's entry instead, e.g. matching on the user agent they care about.

**Compatibility Notes**

### Sitemap-only robots.txt files

The previous implementation emitted a default `*` rule even when a robots.txt file contained only global directives such as `Sitemap:` and no `User-agent` groups.

Example:

```txt
Sitemap: https://example.com/sitemap.xml
```

Previously:

```json
{
  "rules": [
    {
      "user_agent": "*",
      "allow": [],
      "disallow": []
    }
  ]
}
```

Now:

```json
{
  "rules": [],
  "sitemaps": [
    "https://example.com/sitemap.xml"
  ]
}
```

This more accurately reflects the absence of any User-agent rule groups in the source file.

### Directives Before the First User-agent

The previous parser implicitly attached directives encountered before the first `User-agent` to a default `*` rule.

Example:

```txt
Disallow: /private/

User-agent: *
Allow: /
```

With the new rule-group parser, directives that appear before any `User-agent` are ignored because they do not belong to a valid User-agent group.

This behavior aligns with Python's `urllib.robotparser` implementation and the parser's rule-group model.

### Empty User-agent Groups

Completely empty User-agent groups are no longer emitted.

Example:

```txt
User-agent: Googlebot
```

Such groups are skipped unless they contain at least one directive (`Allow`, `Disallow`, or `Crawl-delay`).

Rule groups containing crawler-specific directives such as `Crawl-delay` continue to be preserved.

**Validation**

- [x] `pytest tests/test_robots_txt_tool.py`
- [x] `pytest tests/ -v`
- [x] Tool behaviour tested in `fastmcp inspector` against `inomics.com` and `google.com` 

Closes #110

### DIFF
--- a/tests/test_robots_txt_tool.py
+++ b/tests/test_robots_txt_tool.py
@@ -48,6 +48,78 @@ def test_robots_txt_inspect_happy_path_https(mock_get):
     assert result["rules"][1]["disallow"] == ["/secret/"]
     assert result["rules"][1]["allow"] == []
 
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_separates_groups_across_blank_lines(mock_get):
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "User-agent: Googlebot\n"
+        "Disallow: /private\n"
+        "\n"
+        "User-agent: Bingbot\n"
+        "Allow: /public\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert len(result["rules"]) == 2
+    assert result["rules"][0]["user_agent"] == "Googlebot"
+    assert result["rules"][0]["disallow"] == ["/private"]
+    assert result["rules"][0]["allow"] == []
+    assert result["rules"][1]["user_agent"] == "Bingbot"
+    assert result["rules"][1]["allow"] == ["/public"]
+    assert result["rules"][1]["disallow"] == []
+
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_ignores_trailing_user_agent_without_directives(mock_get):
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "User-agent: *\n"
+        "Disallow: /private\n"
+        "User-agent: Googlebot\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert len(result["rules"]) == 1
+    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["disallow"] == ["/private"]
+    assert result["disallowed_paths"] == ["/private"]
+
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_ignores_directives_before_first_user_agent(mock_get):
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "Disallow: /private\n"
+        "Allow: /public\n"
+        "Crawl-delay: 10\n"
+        "User-agent: *\n"
+        "Disallow: /admin\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert len(result["rules"]) == 1
+    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["disallow"] == ["/admin"]
+    assert result["allowed_paths"] == []
+    assert result["disallowed_paths"] == ["/admin"]
+    assert result["rules"][0]["crawl_delay"] is None
+
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_fallback_to_http(mock_get):
     mock_response = MagicMock()
--- a/tests/test_robots_txt_tool.py
+++ b/tests/test_robots_txt_tool.py
@@ -51,7 +51,7 @@ def test_robots_txt_inspect_happy_path_https(mock_get):
 
 
 @patch("tools.robots_txt_tool.requests.get")
-def test_robots_txt_inspect_separates_groups_across_blank_lines(mock_get):
+def test_robots_txt_inspect_parses_blank_line_separated_rule_groups(mock_get):
     mock_response = MagicMock()
     mock_response.status_code = 200
     mock_response.url = "https://example.com/robots.txt"
--- a/tests/test_robots_txt_tool.py
+++ b/tests/test_robots_txt_tool.py
@@ -1,12 +1,13 @@
-import pytest
 from unittest.mock import patch, MagicMock
 from tools.robots_txt_tool import robots_txt_inspect
 import requests
+
 
 def test_robots_txt_inspect_invalid_domain():
     result = robots_txt_inspect("invalid domain")
     assert result["success"] is False
     assert "Invalid domain format" in result["error"]
+
 
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_happy_path_https(mock_get):
@@ -40,11 +41,11 @@ def test_robots_txt_inspect_happy_path_https(mock_get):
     
     # Check rule sets
     assert len(result["rules"]) == 2
-    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["user_agents"] == ['*']
     assert result["rules"][0]["disallow"] == ["/admin", "/backup/"]
     assert result["rules"][0]["allow"] == ["/admin/public"]
     
-    assert result["rules"][1]["user_agent"] == "Googlebot"
+    assert result["rules"][1]["user_agents"] == ["Googlebot"]
     assert result["rules"][1]["disallow"] == ["/secret/"]
     assert result["rules"][1]["allow"] == []
 
@@ -67,10 +68,10 @@ def test_robots_txt_inspect_separates_groups_across_blank_lines(mock_get):
 
     assert result["success"] is True
     assert len(result["rules"]) == 2
-    assert result["rules"][0]["user_agent"] == "Googlebot"
+    assert result["rules"][0]["user_agents"] == ["Googlebot"]
     assert result["rules"][0]["disallow"] == ["/private"]
     assert result["rules"][0]["allow"] == []
-    assert result["rules"][1]["user_agent"] == "Bingbot"
+    assert result["rules"][1]["user_agents"] == ["Bingbot"]
     assert result["rules"][1]["allow"] == ["/public"]
     assert result["rules"][1]["disallow"] == []
 
@@ -91,7 +92,7 @@ def test_robots_txt_inspect_ignores_trailing_user_agent_without_directives(mock_
 
     assert result["success"] is True
     assert len(result["rules"]) == 1
-    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["user_agents"] == ["*"]
     assert result["rules"][0]["disallow"] == ["/private"]
     assert result["disallowed_paths"] == ["/private"]
 
@@ -114,11 +115,12 @@ def test_robots_txt_inspect_ignores_directives_before_first_user_agent(mock_get)
 
     assert result["success"] is True
     assert len(result["rules"]) == 1
-    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["user_agents"] == ["*"]
     assert result["rules"][0]["disallow"] == ["/admin"]
     assert result["allowed_paths"] == []
     assert result["disallowed_paths"] == ["/admin"]
     assert result["rules"][0]["crawl_delay"] is None
+
 
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_fallback_to_http(mock_get):
@@ -137,6 +139,7 @@ def test_robots_txt_inspect_fallback_to_http(mock_get):
     assert result["disallowed_paths"] == ["/private"]
     assert mock_get.call_count == 2
 
+
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_failure(mock_get):
     mock_get.side_effect = requests.RequestException("Timeout")
@@ -145,6 +148,7 @@ def test_robots_txt_inspect_failure(mock_get):
 
     assert result["success"] is False
     assert "Failed to fetch robots.txt" in result["error"]
+
 
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_parses_crawl_delay_and_host(mock_get):
@@ -166,6 +170,7 @@ def test_robots_txt_inspect_parses_crawl_delay_and_host(mock_get):
     assert result["host"] == "example.com"
     assert result["disallowed_paths"] == ["/private"]
 
+
 @patch("tools.robots_txt_tool.requests.get")
 def test_robots_txt_inspect_crawl_delay_and_host_absent_when_not_present(mock_get):
     mock_response = MagicMock()
@@ -180,8 +185,9 @@ def test_robots_txt_inspect_crawl_delay_and_host_absent_when_not_present(mock_ge
     assert result["host"] is None
     assert result["rules"][0]["crawl_delay"] is None
 
+
 @patch("tools.robots_txt_tool.requests.get")
-def test_robots_txt_inspect_crawl_delay_uses_last_seen_value(mock_get):
+def test_robots_txt_inspect_preserves_per_group_crawl_delay(mock_get):
     mock_response = MagicMock()
     mock_response.status_code = 200
     mock_response.url = "https://example.com/robots.txt"
@@ -220,3 +226,40 @@ def test_robots_txt_inspect_groups_multiple_user_agents(mock_get):
     assert result["rules"][0]["user_agents"] == ["Googlebot", "Bingbot"]
     assert result["rules"][0]["disallow"] == ["/private"]
     assert result["rules"][0]["allow"] == ["/public"]
+
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_crawl_delay_uses_last_seen_value_within_group(mock_get):
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "User-agent: *\n"
+        "Crawl-delay: 5\n"
+        "Crawl-delay: 30\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert result["rules"][0]["crawl_delay"] == "30"
+
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_sitemap_only_file_returns_no_rules(mock_get):
+    mock_response = MagicMock()
+    mock_response.status_code = 200
+    mock_response.url = "https://example.com/robots.txt"
+    mock_response.text = (
+        "Sitemap: https://example.com/sitemap.xml\n"
+    )
+    mock_get.return_value = mock_response
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert result["rules"] == []
+    assert result["sitemaps"] == [
+        "https://example.com/sitemap.xml"
+    ]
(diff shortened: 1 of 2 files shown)


tools/robots_txt_tool.py
```diff
@@ -48,39 +73,51 @@ def robots_txt_inspect(domain: str) -> dict:
...
-                        "disallow": list(dict.fromkeys(current_disallow))
-                    })
-                    current_allow = []
-                    current_disallow = []
-                
-                current_agent = line.split(":", 1)[1].strip()
+                agent = line.split(":", 1)[1].strip()
+
+                # New group starts only after directives have appeared
+                if seen_directive_in_group:
```
SemTiOne, line 82: Blank lines no longer end a rule group.
```
User-agent: Googlebot
User-agent: Bingbot

User-agent: *
Disallow: /
```

`seen_directive_in_group` stays `False` across the blank line, so both UA groups merge into one rule and `Disallow: /` is reported as applying to Googlebot/Bingbot too. Per RFC 9309 a blank line ends a group. Suggest tracking "blank line seen" and starting a new group on the next `User-agent:` line even when no directive appeared (mirroring `urllib.robotparser`'s state machine).

tools/robots_txt_tool.py
```diff
@@ -113,12 +142,11 @@ def robots_txt_inspect(domain: str) -> dict:
             "allowed_paths": list(dict.fromkeys(all_allowed)),
             "disallowed_paths": list(dict.fromkeys(all_disallowed)),
             "sitemaps": list(dict.fromkeys(sitemaps)),
-            "crawl_delay": crawl_delay,
             "host": host,
```
SemTiOne, line 145: The top-level `crawl_delay` field (added in PR #98) is removed from the return dict. Any external consumer reading `result["crawl_delay"]` now gets a `KeyError`, and this is a silent change to the MCP response shape. Either keep the field (last-seen value, as before) or call out the breaking change explicitly in the PR description.

tools/robots_txt_tool.py
```diff
@@ -89,15 +126,7 @@ def robots_txt_inspect(domain: str) -> dict:
...
-        # Add the last rule block if it has anything
-        if current_allow or current_disallow or current_agent == "*":
-            # Avoid adding empty duplicate '*' rules if we haven't seen anything
-            if current_allow or current_disallow or not any(r["user_agent"] == "*" for r in rules):
-                rules.append({
-                    "user_agent": current_agent,
-                    "allow": list(dict.fromkeys(current_allow)),
-                    "disallow": list(dict.fromkeys(current_disallow))
-                })
+        save_rule(current_rule, rules)
```
SemTiOne, line 129: A trailing `User-agent:` with no directives now emits an empty rule (`{user_agent: "Googlebot", allow: [], disallow: [], crawl_delay: None}`) because `save_rule` only checks `user_agents`. The old parser deliberately avoided appending empty rules ("Avoid adding empty duplicate '*' rules"). Same for sitemap-only files, which now return `rules: []` instead of a default `*` rule. Consider skipping rules with no allow/disallow/crawl_delay.

tools/robots_txt_tool.py
```diff
@@ -1,6 +1,29 @@
 import requests
 from utils.helpers import is_valid_domain, normalize_domain
 
+def save_rule(rule: dict, rules: list) -> dict:
+    """Normalize a parsed rule group and append it to the collected rules."""
+    if not rule["user_agents"]:
```
SemTiOne, line 7: Directives appearing before the first `User-agent:` line are now silently dropped: `save_rule` discards the buffered rule because it has no agents, so e.g. `Disallow: /x` followed by `User-agent: *` no longer surfaces `/x` anywhere (not even in `disallowed_paths`). `urllib.robotparser` does the same, so this is defensible, but it's a silent behavior change from the old parser, which attached pre-UA directives to the default `*` rule. Worth being intentional about.

tools/robots_txt_tool.py
```diff
@@ -1,6 +1,29 @@
 import requests
 from utils.helpers import is_valid_domain, normalize_domain
 
+def save_rule(rule: dict, rules: list) -> dict:
```
SemTiOne: `save_rule` returns `None` on every path (bare `return` at line 7 and fall-off-the-end), so the `-> dict` annotation is wrong; it should be `-> None`. The function also mutates `rule`/`rules` in place, which the signature doesn't reflect.

tools/robots_txt_tool.py
```diff
@@ -48,39 +73,51 @@ def robots_txt_inspect(domain: str) -> dict:
...
+                    current_rule = {
+                        "user_agents": [],
+                        "allow": [],
+                        "disallow": [],
+                        "crawl_delay": None,
+                    }
+
+                    seen_directive_in_group = False
+
+                current_rule["user_agents"].append(agent)
```
SemTiOne, line 94: `User-agent:` with an empty value still produces rules with `user_agent: ""`. RFC 9309 says an empty product token should be ignored. Not a regression from the old parser, but worth guarding in `save_rule`, since it now also affects the grouped path (`user_agents: ["", "*"]`).

tests/test_robots_txt_tool.py
```diff
@@ -140,4 +124,27 @@ def test_robots_txt_inspect_crawl_delay_uses_last_seen_value(mock_get):
     result = robots_txt_inspect("example.com")
 
     assert result["success"] is True
-    assert result["crawl_delay"] == "30"
+    assert result["rules"][0]["crawl_delay"] == "5"
+    assert result["rules"][1]["crawl_delay"] == "30"
+
+
+@patch("tools.robots_txt_tool.requests.get")
+def test_robots_txt_inspect_groups_multiple_user_agents(mock_get):
```
SemTiOne, line 210: This test only covers consecutive `User-agent:` lines with no blank line. The three cases most likely to regress are uncovered: blank-line-separated groups, a trailing `User-agent:` with no directives, and directives before the first `User-agent:`.

SemTiOne (changes requested):
Solid work. However, there are some issues (see inline comments above).

tools/robots_txt_tool.py
gaoharimran29-glitch: @SemTiOne Hey,  According to Section 2.1 of RFC 9309, a group is strictly terminated only by a new User-agent line or the end of the file. blank lines are entirely ignored for group termination: "The group is terminated by a user-agent line or end of file." Because blank lines do not split a group, the three User-agent lines legally merge into one group. Therefore, Disallow: / should apply to Googlebot and Bingbot. While legacy parsers like urllib.robotparser (built on 1994 drafts) split on blank lines, modern industry standards including [Google's official open-source parser](https://github.com/google/robotstxt) ignore blank lines and merge these groups exactly as this PR does. The current code is compliant with the RFC.


Resolve it if you agree

tools/robots_txt_tool.py
gaoharimran29-glitch: Good catch, and you're right that this is called out in the PR. I already mentioned moving crawl_delay to per-user-agent groups in the PR. That said, since it's a breaking change to the response shape. Now i had added it more explicit in the PR description itself ("Breaking Changes" section) rather than relying on it being mentioned inline, so it's easy for consumers/reviewers to spot.

tools/robots_txt_tool.py
SemTiOne: @gaoharimran29-glitch You're right on the grouping. I checked RFC 9309 and `emptyline` is explicitly allowed inside a group (`group = startgroupline *(startgroupline / emptyline) *(rule / emptyline)`), and Google's `robots.cc` confirms it: blank lines never touch the state machine, and consecutive `User-agent:` lines only split when a rule line has been seen. So the merge behavior is RFC-compliant and matches Google's parser (current `urllib.robotparser` does the same after the RFC 9309 update in gh-138907). My mistake. You can resolve the comment.

tools/robots_txt_tool.py
gaoharimran29-glitch: Good point. This is a behavior change from the previous implementation. With the new rule-group model, directives encountered before any `User-agent` are ignored because they do not belong to a valid rule group.

This matches the behavior of `urllib.robotparser`, so I left it as-is intentionally. I'll note the behavior change in the PR description `compatibility notes` section for clarity in.

gaoharimran29-glitch:
@SemTiOne Thanks for your review . I addressed your concerns:

What changed:
1. Added logic to filters user-agents with no directive.
2. Updated the type hint in function `save_rule` . Replaced `-> dict` notation with `-> None `.
3. Added logic to filter out user agent with no values.
4. Added three more regression test. They cover the cases you pointed out.
 

Added tests are:

1. Blank-line-separated groups still split into separate rule blocks.
2.  A trailing user-agent with no directives is ignored instead of creating an empty rule.
3. Directives before the first user-agent are discarded and do not leak into the parsed output.

Also updated pr description for the intentional things:
1. Directives encountered before any user-agent are ignored
2. `top_level_crawl_delay` is removed 
3. Sitemap with no default `*` rule even its a only user-directive in robots.txt file

tools/robots_txt_tool.py
```diff
@@ -1,6 +1,34 @@
...
+    """Normalize a parsed rule group and append it to the collected rules."""
+    if not rule["user_agents"]:
+        return
+
+    has_directives = (rule["allow"] or rule['disallow'] or rule['crawl_delay'] is not None)
+
+    if not has_directives:
+        return
+
+    if len(rule["user_agents"]) == 1:
```
SemTiOne: `save_rule` emits `user_agent` (singular) when a rule has exactly one agent but `user_agents` (plural) when it has more than one. Any consumer iterating `result["rules"]` and reading `rule["user_agent"]` will hit a `KeyError` on grouped rules, e.g. `test_robots_txt_inspect_groups_multiple_user_agents`. Could you emit a consistent shape? Either always `user_agents` as a list, or both keys (`user_agent` = first element) for back-compat?

tools/robots_txt_tool.py
```diff
@@ -48,39 +78,54 @@ def robots_txt_inspect(domain: str) -> dict:
...
+                seen_directive_in_group = True
                     
             elif line_lower.startswith("allow:"):
                 path = line.split(":", 1)[1].strip()
+
                 if path:
-                    current_allow.append(path)
+                    current_rule["allow"].append(path)
+
+                seen_directive_in_group = True
```
SemTiOne: `seen_directive_in_group` is flipped even when the directive value is empty (e.g. a bare `Allow:` with no path). That means an empty, no-op line still terminates the group, so a following `User-agent:` starts a fresh group while the previous one is dropped by `save_rule` (no directives). Consider only setting the flag when the value is non-empty (`if path: ... seen_directive_in_group = True`).

tools/robots_txt_tool.py
```diff
@@ -1,6 +1,34 @@
 import requests
 from utils.helpers import is_valid_domain, normalize_domain
 
+def save_rule(rule: dict, rules: list) -> None:
+    """Normalize a parsed rule group and append it to the collected rules."""
+    if not rule["user_agents"]:
+        return
+
+    has_directives = (rule["allow"] or rule['disallow'] or rule['crawl_delay'] is not None)
```
SemTiOne: Minor: mixed quoting (`rule["allow"]` vs `rule['disallow']`), and the `is not None` precedence here makes the parentheses matter. Worth being explicit: `has_directives = bool(rule["allow"] or rule["disallow"]) or rule["crawl_delay"] is not None`. Also a blank line is missing between the imports and `def save_rule` (PEP 8 wants two).

tools/robots_txt_tool.py
```diff
@@ -113,12 +150,11 @@ def robots_txt_inspect(domain: str) -> dict:
...
-            "crawl_delay": crawl_delay,
             "host": host,
             "rules": rules
         }
 
     except requests.RequestException as e:
         return {"success": False, "error": f"Failed to fetch robots.txt: {str(e)}"}
     except Exception as e:
-        return {"success": False, "error": str(e)}
+        return {"success": False, "error": str(e)}
```
SemTiOne: No newline at end of file, add a trailing newline.

tests/test_robots_txt_tool.py
```diff
@@ -140,4 +196,27 @@ def test_robots_txt_inspect_crawl_delay_uses_last_seen_value(mock_get):
     result = robots_txt_inspect("example.com")
 
     assert result["success"] is True
-    assert result["crawl_delay"] == "30"
+    assert result["rules"][0]["crawl_delay"] == "5"
```
SemTiOne, line 205: The name says "uses last seen value", but the test now asserts two different groups (5 and 30). The last-seen-wins semantics within a single group is no longer covered. Rename this (e.g. `...preserves_per_group_crawl_delay`) and add a case with two `Crawl-delay:` lines in the same group.

tests/test_robots_txt_tool.py
```diff
@@ -48,6 +48,78 @@ def test_robots_txt_inspect_happy_path_https(mock_get):
...
+
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert len(result["rules"]) == 1
+    assert result["rules"][0]["user_agent"] == "*"
+    assert result["rules"][0]["disallow"] == ["/admin"]
+    assert result["allowed_paths"] == []
+    assert result["disallowed_paths"] == ["/admin"]
+    assert result["rules"][0]["crawl_delay"] is None
```
SemTiOne, line 122: I suggest a regression test for a sitemap-only file: output changes from a default `*` rule (old parser) to `rules: []` (new parser). That's the most likely output change to surprise consumers, and it's currently untested.

SemTiOne (changes requested):
Thanks for addressing the round-1 points @gaoharimran29-glitch. One item blocks merge (the `user_agent`/`user_agents` shape inconsistency in `save_rule`), everything else is non-blocking polish and can be follow-up.

gaoharimran29-glitch:
@SemTiOne Thanks for the detailed review. I've pushed an update that addresses the feedback raised during review:

* Standardized rule output to always use `user_agents` (list) for a consistent response shape.
* Updated `seen_directive_in_group` handling so that only non-empty directives (Allow, Disallow, Crawl-delay) are treated as group-defining directives. Empty directive values no longer terminate a rule group
* Expanded the `has_rule_directives` condition across multiple lines to improve readability and make the directive checks easier to follow. Now also following consistent double quotes.
* Both files now contains a trailing newline and two lines gap between imports and functions following PEP 8
* Updated tests to reflect the new normalized rule schema, where user_agents is always returned as a list (e.g. ["*"] instead of a singular user_agent: "*") for consistent handling of single-agent and multi-agent rule groups.
* Added a regression test documenting the current behavior for sitemap-only robots.txt files (`rules: []`).
* Added a regression test to verify that when multiple `Crawl-delay` directives appear within the same User-agent rule group, the parser preserves the last-seen value (`Crawl-delay: 30` overrides `Crawl-delay: 5`).
* Renamed the test from `test_robots_txt_inspect_crawl_delay_uses_last_seen_value` to `test_robots_txt_inspect_preserves_per_group_crawl_delay`
* Noticed `pytest` import is unused in the test file so removed it.

tests/test_robots_txt_tool.py
```diff
@@ -40,14 +41,87 @@ def test_robots_txt_inspect_happy_path_https(mock_get):
...
+    result = robots_txt_inspect("example.com")
+
+    assert result["success"] is True
+    assert len(result["rules"]) == 2
+    assert result["rules"][0]["user_agents"] == ["Googlebot"]
+    assert result["rules"][0]["disallow"] == ["/private"]
+    assert result["rules"][0]["allow"] == []
+    assert result["rules"][1]["user_agents"] == ["Bingbot"]
+    assert result["rules"][1]["allow"] == ["/public"]
+    assert result["rules"][1]["disallow"] == []
```
SemTiOne: Misleading name: this test doesn't test blank-line separation. Groups are split by a `User-agent:` line (59-63); the blank line (61) sits mid-group.

SemTiOne (approved):
LGTM! Only 1 minor and non-blocking issue in inline comment below.

tests/test_robots_txt_tool.py
gaoharimran29-glitch: @SemTiOne Good catch. My intention was to cover the common case of rule groups written with blank lines between them (based on the earlier review about blank-line-separated groups), not to imply that blank lines are what trigger group separation in the parser.

I've renamed the test to `test_robots_txt_inspect_parses_blank_line_separated_rule_groups` so it reflects the input being exercised rather than the parser's grouping mechanism.

gaoharimran29-glitch:
@SemTiOne Addressed the remaining non-blocking feedback (renamed the test for clarity) in commit `26fc06c` . Since everything else was already approved, I'm going to merge this. Thanks again for the thorough review!

PR #138, opened 2026-07-31
Improve Nmap scan reliability and result parsing

### LINKED ISSUE #136: Ehancement: Improve Nmap Scan Reliability and Result Parsing
## Summary

Several issues were identified while testing the Nmap-based port scanning functionality across different scan types. Some scan configurations use aggressive timeout values that cause Nmap to terminate scans before completion, resulting in incomplete or empty results. Additionally, OS detection results are currently not exposed to users even when Nmap successfully identifies the target operating system.

### 1. Service Scan Frequently Times Out

Current configuration:

```python
"service": {
    "args": "-sV -F --host-timeout 45s",
    "timeout": 50
}
```

During testing against public targets (e.g. `google.com`), service detection required significantly longer than 45 seconds to complete. Nmap terminated the scan due to the host timeout before service/version information could be collected.

Observed behavior:

```json
{
  "host": "...",
  "state": "up",
  "protocols": {}
}
```

Expected behavior:

```json
{
  "port": 80,
  "service": "http",
  "product": "...",
  "version": "..."
}
```

### 2. OS Scan Does Not Return OS Information

Current configuration:

```python
"os": {
    "args": "-O -F --host-timeout 45s",
    "timeout": 50
}
```

The scan executes successfully, but the implementation only returns port information.

Testing against `scanme.nmap.org` showed that Nmap successfully generated OS fingerprints:

```text
Aggressive OS guesses:
Linux 4.19 - 5.15 (98%)
Linux 4.15 (94%)
...
```

However, the tool response only includes open ports.

Current output:

```json
{
  "host": "...",
  "protocols": {
    "tcp": [...]
  }
}
```

Expected output:

```json
{
  "host": "...",
  "os_matches": [
    {
      "name": "Linux 4.19 - 5.15",
      "accuracy": "98"
    }
  ],
  "protocols": {
    "tcp": [...]
  }
}
```

The implementation should extract and return OS detection data (e.g. `osmatch`) when `scan_type="os"`.

### 3. Full Scan Can Be Aborted Prematurely

Current configuration:

```python
"full": {
    "args": "-p- --host-timeout 5m",
    "timeout": 100
}
```

Testing showed that full port scans against Internet-facing hosts may take several minutes. Previous configurations using `--host-timeout` caused Nmap to terminate scans before completion.

A suitable timeout strategy should be defined for full scans to avoid incomplete results while still preventing scans from running indefinitely.

### 4. Vulnerability Scan Requires Longer Execution Time

Current configuration:

```python
"vuln": {
    "args": "--script vuln -F"
    "timeout": 100
}
```

Testing against `scanme.nmap.org` showed vulnerability scans taking over 5 minutes to complete.

Example:

```text
Nmap done: 1 IP address (1 host up) scanned in 325.45 seconds
```

Current timeout values are insufficient for many NSE vulnerability scripts and can cause scans to terminate before results are returned.

## Proposed Changes

* Increase timeout values for `service`, `full`, and `vuln` scan types. ( Add suffiicient timeout by testing so most of the website gets scanned successfully )
* Re-evaluate use of aggressive `--host-timeout` values.
* Return OS fingerprinting results (`osmatch`) for OS scans.
* Consider returning scan duration in the response for easier debugging and tuning.

## Acceptance Criteria

* Service scans successfully return version information for targets that support service detection.
* OS scans include OS match data and accuracy scores when available.
* Full and vulnerability scans have timeout settings appropriate for long-running operations.
* Scan results contain enough metadata to distinguish successful scans from partial results.

Files to be touched : - `tools/portscan_tool.py` , `tests/test_port_scan.py`

SemTiOne:
Closes #136

## Changes

- Relaxed `--host-timeout`/`timeout` for `service`, `os`, `full`, and `vuln` (with `--script-timeout 5m` on `vuln`), so long scans no longer terminate early.
- `os` scans now return `os_matches` (`{name, accuracy}`); results also include `duration_seconds` and `host_timeout_status` (keyed by IP) to distinguish complete scans from partial ones.

## Tests

Expanded `tests/test_port_scan.py` covering the new configs and parsin logic, all pass.

### DIFF
--- a/tools/portscan_tool.py
+++ b/tools/portscan_tool.py
@@ -1,15 +1,36 @@
+import re
 import xml.etree.ElementTree as ET
 import nmap
 
 SCAN_CONFIG = {
-    "basic": {"args": "-F --host-timeout 30s", "timeout": 35},
-    "service": {"args": "-sV -F --host-timeout 120s", "timeout": 130},
-    "os": {"args": "-O -F --host-timeout 60s", "timeout": 65},
-    "full": {"args": "-p- --host-timeout 15m", "timeout": 950},
-    "vuln": {"args": "--script vuln -F --host-timeout 15m --script-timeout 5m", "timeout": 950},
+    "basic": {"args": "-F --host-timeout 30s"},
+    "service": {"args": "-sV -F --host-timeout 120s"},
+    "os": {"args": "-O -F --host-timeout 60s"},
+    "full": {"args": "-p- --host-timeout 15m"},
+    "vuln": {"args": "--script vuln -F --host-timeout 15m --script-timeout 5m"},
 }
 
+TIMEOUT_MARGIN = 0.2
+
 PORT_SCAN_TIMEOUT_ERRORS = (TimeoutError, getattr(nmap, "PortScannerTimeout", TimeoutError))
+
+_DURATION_UNITS = {"s": 1, "m": 60, "h": 3600}
+
+
+def _host_timeout_seconds(args: str) -> int:
+    match = re.search(r"--host-timeout\s*[=]?\s*(\d+)\s*([smh])?", args)
+    if not match:
+        return 0
+    value = int(match.group(1))
+    unit = match.group(2) or "s"
+    return value * _DURATION_UNITS[unit]
+
+
+def _default_timeout(args: str) -> int:
+    host_timeout = _host_timeout_seconds(args)
+    if host_timeout <= 0:
+        raise ValueError(f"No parseable --host-timeout in args: {args!r}")
+    return int(host_timeout * (1 + TIMEOUT_MARGIN))
 
 
 def _extract_host_timeout_status(nmap_output: bytes) -> dict:
@@ -37,7 +58,7 @@ def _extract_host_timeout_status(nmap_output: bytes) -> dict:
     return result
 
 
-def port_scan(target: str, scan_type: str = "basic") -> dict:
+def port_scan(target: str, scan_type: str = "basic", timeout: int | None = None) -> dict:
     """
     Perform Nmap port scan on a target IP or domain.
 
@@ -47,6 +68,9 @@ def port_scan(target: str, scan_type: str = "basic") -> dict:
     - "os"      : OS detection, needs admin (-O -F)
     - "full"    : All 65535 ports, slow (-p-)
     - "vuln"    : Basic vulnerability scripts (--script vuln -F)
+
+    timeout overrides the application-level timeout in seconds. When omitted,
+    it defaults to the scan's --host-timeout plus a margin (TIMEOUT_MARGIN).
     """
     if scan_type not in SCAN_CONFIG:
         return {
@@ -58,14 +82,21 @@ def port_scan(target: str, scan_type: str = "basic") -> dict:
             "valid_scan_types": list(SCAN_CONFIG.keys()),
         }
 
+    if timeout is not None and timeout < 1:
+        return {
+            "success": False,
+            "error": f"Invalid timeout '{timeout}'. Must be a positive number of seconds.",
+        }
+
     try:
         scanner = nmap.PortScanner()
 
         config = SCAN_CONFIG[scan_type]
+        effective_timeout = timeout if timeout is not None else _default_timeout(config["args"])
         scanner.scan(
             hosts=target,
             arguments=config["args"],
-            timeout=config["timeout"],
+            timeout=effective_timeout,
         )
 
         results = []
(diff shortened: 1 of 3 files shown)


tools/portscan_tool.py
```diff
@@ -50,6 +77,14 @@ def port_scan(target: str, scan_type: str = "basic") -> dict:
                 "protocols": {}
             }
 
+            os_matches = scanner[host].get("osmatch", [])
+            if os_matches:
+                host_data["os_matches"] = [
+                    {"name": m["name"], "accuracy": m["accuracy"]}
+                    for m in os_matches
+                    if m.get("name") and m.get("accuracy")
```
gaoharimran29-glitch: Why are we filtering out matches that don't contain both name and accuracy? This could hide potentially useful fingerprint information returned by Nmap. Could we keep the entry and expose whatever fields are available instead?

tools/portscan_tool.py
```diff
@@ -50,6 +77,14 @@ def port_scan(target: str, scan_type: str = "basic") -> dict:
                 "protocols": {}
             }
 
+            os_matches = scanner[host].get("osmatch", [])
```
gaoharimran29-glitch, line 111: Issue description specifically asked to expose osmatch only for OS scans. Add a if conditon like `if scan_type == "os" and os_matches:`

tools/portscan_tool.py
```diff
@@ -1,15 +1,42 @@
...
 SCAN_CONFIG = {
     "basic": {"args": "-F --host-timeout 30s", "timeout": 35},
-    "service": {"args": "-sV -F --host-timeout 45s", "timeout": 50},
-    "os": {"args": "-O -F --host-timeout 45s", "timeout": 50},
-    "full": {"args": "-p- --host-timeout 5m", "timeout": 310},
-    "vuln": {"args": "--script vuln -F --host-timeout 90s", "timeout": 100},
+    "service": {"args": "-sV -F --host-timeout 120s", "timeout": 130},
+    "os": {"args": "-O -F --host-timeout 60s", "timeout": 65},
+    "full": {"args": "-p- --host-timeout 15m", "timeout": 950},
+    "vuln": {"args": "--script vuln -F --host-timeout 15m --script-timeout 5m", "timeout": 950},
```
gaoharimran29-glitch: One concern I still have is how the timeout values were chosen. Scan duration varies significantly across targets and even across runs of the same scan type. For example, I've observed while testing `scanme.nmap.org` vuln scans ranging from a few minutes to ~970 seconds.

Additionally, should we consider making the timeout configurable by the caller (with sensible defaults provided by the tool)? That would allow users to increase the limit for slower targets while still preserving a safe default experience for typical scans.

gaoharimran29-glitch (commented):
@SemTiOne Thanks for the contribution and for adding the additional tests. The OS detection support and scan metadata (duration_seconds, host_timeout_status) are useful improvements, and I appreciate the effort put into expanding the test coverage.

I've left a few review comments/questions on the PR. When you get a chance, please take a look. Looking forward to your response.

tools/portscan_tool.py
SemTiOne: I agree. I'll expose whatever fields are available ({k: m.get(k) for k in ("name", "accuracy")}) instead of dropping incomplete entries.

tools/portscan_tool.py
SemTiOne: Good catch. Your ~970s observation shows hardcoded values can still kill legit slow scans (950s limit < 970s). I'll make the app-level timeout configurable (`port_scan(..., timeout=None)`) with the default derived from `--host-timeout` + margin, so callers can raise it for slow targets.

SemTiOne:
@gaoharimran29-glitch Thanks for the review, all three points are addressed in `b316687`:

- **`os_matches`**: now limited to `os` scans only (`portscan_tool.py:112`), and keeps partial entries (`{k: m.get(k) ...}`) instead of dropping them, so `full_recon` stays intact.
- **Timeout**: app-level timeout is now derived from `--host-timeout` (+20% margin) and overridable via a new `timeout` param, so no more hardcoded 950s < 970s issue.
- **`duration_seconds` / `host_timeout_status`** added, keyed by IP with IPv6 fallback, consistent with how `all_hosts()` is keyed, so callers can distinguish complete vs partial scans.

Also `mcp.json` updated.

gaoharimran29-glitch (approved):
Thanks for taking the time to address the review comments and iterate on the implementation.

The latest changes look good to me. Thanks again for your contribution!

Issue #130, opened 2026-07-23
Refactor: tech_stack_detect into a Dedicated Technology Fingerprinting Engine


gaoharimran29-glitch:
## Summary

The current `tech_stack_detect` tool provides basic technology fingerprinting but suffers from several accuracy, architecture, and maintenance issues. It currently mixes technology detection with security header analysis, relies heavily on simple string matching, produces misleading classifications in some cases, and lacks modern fingerprinting techniques.

This issue proposes refactoring the tool into a dedicated technology fingerprinting engine focused exclusively on identifying technologies used by a target website.

---

### 1. Responsibility Overlap with `headers_analyzer`

The tool currently performs security header analysis and generates security scores:

* Security header presence checks
* Security score calculation
* Security rating generation

However, this functionality already exists in the `headers_analyzer` tool.

This results in:

* Duplicate logic
* Inconsistent results across tools
* Additional maintenance burden
* Technology detection results being mixed with security assessments

### 2. Inaccurate Security Ratings

The current scoring mechanism often produces misleading results.

Examples observed during testing:

* Google homepage reported as "Poor"
* OpenAI reported as "Fair"

These ratings do not accurately represent the real-world security posture of those services.

The issue stems from:

* Equal weighting of all headers
* Reliance on obsolete headers such as `X-XSS-Protection`
* Lack of contextual interpretation

Since header analysis belongs in `headers_analyzer`, this functionality should be removed entirely.

### 3. Weak Technology Detection Logic

Current detection is based primarily on:

```python
if signature in html
```

This approach causes:

* False positives
* False negatives
* Missed modern frameworks
* Limited confidence in results

Examples:

* React detection frequently misses modern React applications
* Next.js sites may be detected without React
* Tailwind detection is unreliable
* Framework detection depends heavily on exposed asset names

### 4. Misleading Server Identification

The tool currently exposes:

```json
{
  "web_server": "cloudflare"
}
```

or

```json
{
  "web_server": "github.com"
}
```

These values are often raw `Server` header values and do not necessarily represent actual web server software.

Examples:

* Cloudflare is an edge provider/reverse proxy
* github.com is a custom server header
* gws represents Google Web Server

The current field name can therefore be misleading.

### 5. Missing Detection Layers

Technology identification currently relies almost entirely on:

* Response headers
* HTML source

Important fingerprinting sources are not used:

* Cookies
* Meta generator tags
* Script URLs
* Asset URLs
* Framework-specific markers
* Platform-specific headers

[...]

Issue #140, opened 2026-07-31
List-valued technologies reach the LLM prompt as a Python repr, e.g. Software detected : nginx, ['WordPress']


Nitjsefnie:
### Describe the bug

When a detected technology is a list, `techstack_extractor` stringifies the Python list object, so a literal `['WordPress']` — brackets and quotes included — is fed into the LLM threat-analysis prompt.

`tools/signals/tech_stack.py:10-12`:

```python
for tech in technologies.values():
    if tech and str(tech).strip() not in ("Unknown", "None", ""):
        signals["software_detected"].append(str(tech).strip())
```

`tech_stack_detect` stores lists for several categories — `tech["cdn"]`, `tech["cms"]`, the JS-framework list and the analytics list are all built as lists in `tools/techstack_tool.py` — so `str(tech)` yields the repr rather than the names.

Reproduced end-to-end on `main` with a realistic nginx + WordPress result, running the real extractor and the real prompt formatter:

```
software_detected = ['nginx', "['WordPress']"]
PROMPT LINE >>>Software detected  : nginx, ['WordPress']<<<
```

The model is therefore told the detected software is `['WordPress']`. A site with two CMS matches would read `['WordPress', 'Drupal']` as a single item.

### To Reproduce

1. Take any `tech_stack_detect` result whose `technologies` contains a list value — e.g. a WordPress site, which sets `cms` to a list.
2. Run `techstack_extractor(result, signals)` and inspect `signals["software_detected"]`.
3. The list-valued entries appear as their Python repr, and carry through to the prompt's `Software detected` line.

### Expected behavior

List-valued technologies are flattened into individual names, so the prompt reads `Software detected : nginx, WordPress`.

### Additional context

Found while doing #130 Phase 1. It is in the signal extractor rather than the tool, and independent of that refactor, so I have not touched it there — the extractor is explicitly out of scope for Phase 1.

Unverified suggestion: iterating list values rather than stringifying them, and appending each element, would produce the intended output; scalars would keep their current behaviour.

gaoharimran29-glitch:
@Nitjsefnie Thanks for spotting this and taking the initiative to fix it! Really appreciate that you found the bug, opened the issue, and followed through with a PR , that's exactly the kind of contribution that helps keep the project improving.

Issue #139, opened 2026-07-31
full_recon can never get security headers from headers_analyzer: it is not in the signal registry


Nitjsefnie:
### Describe the bug

`full_recon`'s `missing_security_headers` signal can never be populated by `headers_analyzer`, because `headers_analyzer` is not in the signal registry at all — so `full_recon` never runs it.

Verified on `main`:

- `tools/signals/registry.py` — `TOOL_REGISTRY` has exactly nine entries: `whois`, `dns`, `ssl`, `email_security`, `asn`, `ports`, `techstack`, `ct_logs`, `ip_reputation`. `grep -c "headers"` over that file returns **0**.
- The only two writes to `signals["missing_security_headers"]` anywhere in the tree are `tools/signals/tech_stack.py:21` and `:26` — i.e. the signal has only ever come from `tech_stack_detect`.
- `tools/signals/extractor.py:15` initialises the default and carries the trailing comment `# headers`, which reads like a headers extractor was intended and never registered.
- `headers_analyzer` is wired only as a standalone MCP tool (`server.py`, `mcp.json`), never into the registry, `fullrecon_tool`, or any extractor.

This matters because the threat-analysis prompt tells the model the signal comes from techstack, in three places (`tools/prompts/threat_analysis.py`):

```
:11  • Missing security headers (techstack) + exposed web services = increased browser attack surface
:16  • Many CT-log subdomains (ct_logs) + missing security headers (techstack) = broad attack surface
     with weak baseline hardening
:84  • IF the techstack scan failed, returned an HTTP status of 4xx/5xx, or no security headers were
     collected, classify security headers as "Insufficient data"…
```

So the header signal's only source is the tool that #130 says should not be doing header analysis, and the tool that should be doing it is unreachable from `full_recon`.

### To Reproduce

1. On `main`, `grep -n '"name"' tools/signals/registry.py` — nine entries, no `headers`.
2. `grep -rn 'missing_security_headers' tools/` — only `tech_stack.py:21`, `tech_stack.py:26` write it; `extractor.py:15` initialises it; `fullrecon_tool.py:26` reads it.
3. `grep -rn 'headers_analyzer' tools/signals/ tools/fullrecon_tool.py` — no hits.

### Expected behavior

`full_recon`'s security-header signal comes from `headers_analyzer` — the tool that owns header analysis — rather than from `tech_stack_detect`; or, if that wiring is not wanted, the threat-analysis prompt does not name a source that cannot supply it.

### Additional context

[...]

PR #141, opened 2026-07-31
refactor(techstack): remove security-header analysis from tech_stack_detect (#130 phase 1)

Nitjsefnie:
## Summary

Phase 1 of #130: `tech_stack_detect` no longer does security-header analysis. The header list, the present/missing split and the `security_score`/`rating` computation are gone, along with the `security_headers` key in the returned dict. Everything else in the tool is byte-for-byte unchanged, including all three `except` clauses.

Scoped to Phase 1 deliberately — Phases 2–4 and the proposed output schema are left for follow-ups so this stays reviewable. Say the word if you would rather have it in one go.

## Related Issues and Pull Requests

Partially addresses #130 — this is phase 1 only; the remaining phases follow in separate PRs.

## Changes

- `tools/techstack_tool.py` — removed the `# ── Security Headers ──` block and the `"security_headers": {...}` entry; docstring now reads "...CDN, and analytics".
- `tests/test_tech_stack_detect.py` — two tests removed (below), one added.
- `README.md`, `mcp.json` — dropped the security-header mentions from the tool's description. `server.json` has no per-tool entries, so nothing to change there.

The output schema is the old one minus `security_headers`; the issue's proposed schema is **not** adopted here, matching how increments have landed in this repo before. `tools/signals/tech_stack.py` is untouched.

## Testing

`pytest tests/ -v` — the CI job's exact command. **288 passed / 15 subtests** on `main`, **287 / 15** here: −2 deleted, +1 added. No skips, no xfails, no relaxed assertions; only `tests/test_tech_stack_detect.py` was touched.

The new `test_no_security_analysis_in_output` was written **before** the change and run against unmodified `tools/techstack_tool.py`, where it fails:

```
AssertionError: 'security_headers' unexpectedly found in {..., 'security_headers':
{'present': [...], 'missing': [...], 'score': '28%', 'rating': 'Poor'}}
```

It feeds headers that *are* present, so the old code produced a non-trivial block rather than an empty one, and it pins the whole top-level key set against a pasted literal — so re-introducing the behaviour under a different key name fails it too.

**On the two deleted tests**, since that is the part worth scrutinising:

- `test_security_header_scoring` — asserted only `result["security_headers"]["score"] == "100%"`, `["rating"] == "Excellent"`, `["missing"] == []`
- `test_poor_security_header_rating` — asserted only `["score"] == "0%"`, `["rating"] == "Poor"`

Every assertion in both reads `result["security_headers"]`; neither has any residual fingerprinting assertion that could have been kept. To be precise about the authority: the issue does not mention tests at all — it says this functionality "should be removed entirely", and these tests assert exactly that functionality, so they go with it and are replaced by a test pinning its absence.

No other test asserts on the tool's `security_headers`. The other hits in `tests/` (`test_signals_extractors.py`, `test_asn_extractor.py`, `test_full_recon.py`) hand-build synthetic `signals` dicts and never call `tech_stack_detect` — and the `full_recon` tests patch both `TOOL_REGISTRY` and `extract_signals`, so the real tool and extractors never run. `tests/test_headers_tool.py` is untouched and keeps the genuine header coverage.

## Bugs Discovered

- #139 full_recon can never get security headers from headers_analyzer: it is not in the signal registry
- #140 List-valued technologies reach the LLM prompt as a Python repr, e.g. Software detected : nginx, ['WordPress']

## Follow-ups / Known Limitations

**The disclosed consequence, stated fully because it is slightly worse than it first looks.** With the security block gone, `techstack_extractor` takes its `not headers_data` path on every successful run, so `full_recon`'s `missing_security_headers` becomes permanently `[]` and renders as `Missing sec headers: 0 — none`. I verified that end to end rather than reasoning about it: no crash, the extractor's graceful absent-data path handles it, no auto-warnings.

Beyond what I said when claiming the issue: `tools/prompts/threat_analysis.py` names `(techstack)` as that signal's source in three places, and its "Insufficient data" rule keys on the techstack scan — so after this PR the prompt describes a source that produces nothing, and that rule can never fire. The underlying cause is that `headers_analyzer` was never registered, which is #139.

I have deliberately **not** migrated the signal here, because that is registry wiring rather than this refactor and it is your call. If you would rather it landed together, say so and I will fold it in.

Two smaller notes so they are not "you missed a spot": `tools/signals/test.py` still carries an old sample payload with `security_headers`, left alone as non-executed sample data that pytest does not collect; and the branch is cut from `main` rather than my fork's default, which was 20 commits behind.

Generated by Claude Opus 5 (brief, implementation, review), Kimi K3 (issue triage, verification)

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ It is also listed on glama mcp registry.
 | `port_scan` | Nmap-powered scanner with service/version detection and security warnings |
 | `ssl_inspect` | SSL/TLS certificate — issuer, expiry, cipher strength, SANs, TLS version |
 | `email_security_check` | Checks SPF, DKIM, and DMARC DNS records — returns a security_score, rating, and actionable recommendations for missing or weak configurations |
-| `tech_stack_detect` | Web server, CMS, JS frameworks, CDN, analytics, and security header scoring |
+| `tech_stack_detect` | Web server, CMS, JS frameworks, CDN, and analytics |
 | `cert_transparency` | Query crt.sh Certificate Transparency logs for a domain and extract certificate, subdomain, and wildcard information |
 | `asn_lookup` | Autonomous System Number (ASN) and network ownership lookup via Team Cymru WHOIS — identifies ASN, BGP prefix, organization, registry, country, and allocation date for domains or IP addresses (no API key required) |
 | `ip_reputation` | Check if an IP is flagged as malicious via AbuseIPDB (api key requied) |
--- a/mcp.json
+++ b/mcp.json
@@ -68,7 +68,7 @@
     },
     {
       "name": "tech_stack_detect",
-      "description": "Fingerprint web server, CMS, CDN, JS frameworks, and security headers",
+      "description": "Fingerprint web server, CMS, CDN, and JS frameworks",
       "input": {
         "domain": "string"
       },
--- a/tests/test_tech_stack_detect.py
+++ b/tests/test_tech_stack_detect.py
@@ -51,30 +51,23 @@ class TestTechStackDetect(unittest.TestCase):
         self.assertIn("Next.js", result["technologies"]["javascript_frameworks"])
 
     @patch("tools.techstack_tool.requests.get")
-    def test_security_header_scoring(self, mock_get):
+    def test_no_security_analysis_in_output(self, mock_get):
         headers = {
+            "server": "nginx/1.18",
             "strict-transport-security": "max-age=31536000",
             "x-frame-options": "DENY",
-            "x-content-type-options": "nosniff",
-            "content-security-policy": "default-src 'self'",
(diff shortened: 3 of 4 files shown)


gaoharimran29-glitch (changes requested):
@Nitjsefnie Looks good to me. One request before merging: since this intentionally implements only Phase 1 of #130 and leaves the remaining phases for follow-up PRs, could you update the PR description to replace Fixes #130 with Partially addresses #130 (or similar) ? That will prevent GitHub from automatically closing the issue.

Nitjsefnie:
Done — the body now says `Partially addresses #130` instead, so the issue will stay open for the remaining phases. Good catch.

gaoharimran29-glitch:
@Nitjsefnie Thanks for the quick update. Its merged now

PR #158, opened 2026-08-13
fix(signals): register headers_analyzer so full_recon can report missing security headers

### LINKED ISSUE #139: full_recon can never get security headers from headers_analyzer: it is not in the signal registry

Nitjsefnie:
## Closes

Fixes #139

Relates to #130 — not closed here; this only removes the dead half its merged phase 1 (#141) left behind.

## Summary

`headers_analyzer` was never in `TOOL_REGISTRY`, so once #141 removed header analysis from `tech_stack_detect` nothing could populate `missing_security_headers` and `full_recon` reported "0 — none" for every domain. This registers `headers` as a wave-2 tool with its own signal extractor.

## What changed

- New `tools/signals/headers.py`: `headers_extractor` builds `missing_security_headers` from `headers_analyzer`'s per-header `present` flags, and keeps the two existing auto-warning tiers (>= 4 missing → "significant hardening gap", >= 2 → plain listing).
- The signal is only populated when the hop that was actually analysed answered 2xx. `headers_analyzer` analyses whatever the final redirect hop served, and an error or block page's own bare headers say nothing about the site — that is insufficient data, not a hardening finding. Failed, empty-chain and malformed results leave the signal at its default.
- `tools/signals/registry.py`: `headers` registered in wave 2, next to `techstack`.
- `tools/signals/tech_stack.py`: dropped the block still reading `result["security_headers"]` / `result["status_code"]`, keys `tech_stack_detect` stopped returning in #141.
- `tools/prompts/threat_analysis.py`: the header signal is attributed to `(headers)` instead of `(techstack)`, and the "Insufficient data" rule now also covers a bot/WAF-blocked scan and a scan that collected no headers.
- `README.md` / `CONTRIBUTING.md`: `headers_analyzer` moved to the tools `full_recon` runs, and `headers` added to the wave-2 list.

## Notes

- The 2xx cutoff is narrower than it looks: `headers_analyzer` already returns `success: False` for a chain ending on a redirect and for a detected WAF/bot page, so in practice it suppresses the signal on 4xx/5xx only.
- Names in the signal are `headers_analyzer`'s own lowercase HTTP header names (`content-security-policy`, `x-frame-options`, …), so they appear in that spelling in the auto-warning text too.

## Verification

- [x] Ran locally and confirmed it works as expected — focused runs of the extractor across the status boundary (199/200/299/300/304/403/500, plus empty and malformed redirect chains) and of `full_recon` with only the HTTP hop canned.
- [x] Added/updated tests for the changed behavior — `tests/test_signals_extractors.py`, `tests/test_full_recon.py`, and a new `tests/test_threat_analysis_prompt.py` pinning the prompt's evidence rule to the extractor's cutoff.
- [x] All tests pass (`pytest tests/ -v`) — 329 passed, 15 subtests passed, in Actions run 31688507644 on our fork Nitjsefnie-OSC/AynOps. That run's log records `checked-out SHA: ccb4e1a32dbd1d32c634037539293caece23c11d`, this branch's head; the run's own `head_sha` is the fork's default branch, not the tree tested.
- [x] Checked `CONTRIBUTING.md` and applied all changes — its wave-2 tool list now includes `headers`.
- [x] Updated Documentation ( if required ) — `README.md` tool tables.

Generated by GPT-5.6 Sol (brief, review), Claude Opus 5 (implementation, verification, testing, fork CI)

### DIFF
--- a/tools/signals/headers.py
+++ b/tools/signals/headers.py
@@ -1,0 +1,34 @@
+def headers_extractor(result, signals):
+    """Populate the security-header signal from a headers_analyzer result.
+
+    headers_analyzer reports every checked header as
+    ``{"present": bool, "value": ..., "issue": ..., "severity": ...}``;
+    the ones it marks absent are what `missing_security_headers` means.
+    Information-disclosure headers (``server``, ``x_powered_by``, …) only
+    appear when they ARE present, so they never enter the missing list.
+    """
+    if not result.get("success"):
+        return
+
+    headers = result.get("headers")
+
+    if not isinstance(headers, dict):
+        return
+
+    missing = [
+        name
+        for name, analysis in headers.items()
+        if isinstance(analysis, dict) and not analysis.get("present")
+    ]
+
+    signals["missing_security_headers"] = missing
+
+    if len(missing) >= 4:
+        signals["auto_warnings"].append(
+            f"{len(missing)} security headers missing ({', '.join(missing)}) — significant hardening gap"
+        )
+
+    elif len(missing) >= 2:
+        signals["auto_warnings"].append(
+            f"{len(missing)} security headers missing: {', '.join(missing)}"
+        )
--- a/tools/signals/headers.py
+++ b/tools/signals/headers.py
@@ -1,3 +1,28 @@
+def _final_hop_status(result):
+    """The status code of the response headers_analyzer analysed, or None.
+
+    headers_analyzer does not surface the status at the top level; the last
+    entry of ``redirect_chain`` is the hop whose headers became the
+    analysis.
+    """
+    chain = result.get("redirect_chain")
+
+    if not isinstance(chain, list) or not chain:
+        return None
+
+    final_hop = chain[-1]
+
+    if not isinstance(final_hop, dict):
+        return None
+
+    status = final_hop.get("status_code")
+
+    if isinstance(status, bool) or not isinstance(status, int):
+        return None
+
+    return status
+
+
 def headers_extractor(result, signals):
     """Populate the security-header signal from a headers_analyzer result.
 
@@ -6,8 +31,20 @@ def headers_extractor(result, signals):
     the ones it marks absent are what `missing_security_headers` means.
     Information-disclosure headers (``server``, ``x_powered_by``, …) only
     appear when they ARE present, so they never enter the missing list.
+
+    A result that did not reach real page content is insufficient data, not
+    a hardening finding: headers_analyzer happily analyses whatever the
+    final hop returned, and an error, maintenance or bot-block page carries
+    its own (usually bare) headers, which say nothing about the site. Unless
+    the analysed hop answered 2xx, the signal is left at its default so the
+    prompt's "Insufficient data" rule applies instead.
     """
     if not result.get("success"):
+        return
+
+    status = _final_hop_status(result)
+
+    if status is None or not 200 <= status < 300:
         return
 
     headers = result.get("headers")
--- a/tools/signals/headers.py
+++ b/tools/signals/headers.py
@@ -1,29 +1,4 @@
-def _final_hop_status(result):
-    """The status code of the response headers_analyzer analysed, or None.
-
-    headers_analyzer does not surface the status at the top level; the last
-    entry of ``redirect_chain`` is the hop whose headers became the
-    analysis.
-    """
-    chain = result.get("redirect_chain")
-
-    if not isinstance(chain, list) or not chain:
-        return None
-
-    final_hop = chain[-1]
-
-    if not isinstance(final_hop, dict):
-        return None
-
-    status = final_hop.get("status_code")
-
-    if isinstance(status, bool) or not isinstance(status, int):
-        return None
-
-    return status
-
-
-def headers_extractor(result, signals):
+def headers_extractor(result, signals) -> None:
     """Populate the security-header signal from a headers_analyzer result.
 
     headers_analyzer reports every checked header as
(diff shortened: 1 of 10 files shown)


gaoharimran29-glitch:
Hey @SemTiOne , could you please review this PR when you get a chance? I have a couple of other PRs to go through as well and I’m a bit tied up at the moment, so I’m assigning this one to you. Would really appreciate you taking a look and sharing your feedback. Thanks! 🙌

tools/signals/headers.py
```diff
@@ -0,0 +1,71 @@
...
+        return None
+
+    status = final_hop.get("status_code")
+
+    if isinstance(status, bool) or not isinstance(status, int):
+        return None
+
+    return status
+
+
```
SemTiOne: Delete this function.

tools/signals/headers.py
```diff
@@ -0,0 +1,71 @@
...
+    the analysed hop answered 2xx, the signal is left at its default so the
+    prompt's "Insufficient data" rule applies instead.
+    """
+    if not result.get("success"):
+        return
+
+    status = _final_hop_status(result)
+
+    if status is None or not 200 <= status < 300:
+        return
```
SemTiOne, line 30:
```suggestion
     chain = result.get("redirect_chain")
     if not isinstance(chain, list) or not chain:
         return
     final_hop = chain[-1]
     if not isinstance(final_hop, dict):
         return
     status = final_hop.get("status_code")
     if isinstance(status, bool) or not isinstance(status, int):
         return
     if not 200 <= status < 300:
         return
```

tools/signals/headers.py
```diff
@@ -0,0 +1,71 @@
...
+
+    status = final_hop.get("status_code")
+
+    if isinstance(status, bool) or not isinstance(status, int):
+        return None
+
+    return status
+
+
+def headers_extractor(result, signals):
```
SemTiOne:
```suggestion
def headers_extractor(result, signals) -> None:
```

SemTiOne (approved):
Nice work! A few minor tweaks and let's merge.

Three nits:
1. Delete unused `_final_hop_status` helper (lines 1-25)
2. Add `-> None` return type on `headers_extractor`
3. Move `isinstance(status, bool)` check before range check

See below.

tools/signals/headers.py
Nitjsefnie: Removed `_final_hop_status` in `b9a7c9f`; the extractor now performs the same validation inline.

tools/signals/headers.py
Nitjsefnie: Updated in `b9a7c9f`: booleans and non-integers are rejected before the 2xx range check.

tools/signals/headers.py
Nitjsefnie: Added the `-> None` return annotation in `b9a7c9f`.

gaoharimran29-glitch:
@SemTiOne Hey , just a quick reminder to review the latest changes on this PR whenever you get a chance. Thanks!

SemTiOne (approved):
LGTM!

Issue #154, opened 2026-08-10
Enhancement: Improve Subdomain Takeover Detection Accuracy & Reliability


gaoharimran29-glitch:
# Summary

Current flow works but has correctness gaps that cause false positives/negatives and lose useful error info. This issue is correctness/reliability only — no rewrite, no new services unless needed to fix fingerprint architecture.

**Contributing:** Items are independent. Comment with the number(s) you want before starting, to avoid duplicate work. Include tests for whatever you change.

---

## A. DNS handling

- [x] **1. Don't treat DNS failures as "safe."**
`_resolve_cname()` catches `Exception` and returns `None` for *everything* — no CNAME, NXDOMAIN, timeout, resolver error. Caller then marks all of these `safe`. Fix: return a structured result (or enum) that distinguishes `no_cname` / `dns_error` / `cname_found`. Catch specific `dnspython` exceptions (`NoAnswer`, `NXDOMAIN`, timeouts) instead of bare `except Exception`; don't swallow unexpected errors.

- [x] **2. Follow the full CNAME chain.**
Only the first CNAME hop is checked, so `blog.example.com → alias.example.net → something.github.io` never reaches the real target. Fix: resolve iteratively up to a max depth (e.g. 5), detect loops, and keep the full chain for evidence/debugging.

## B. Result classification

- [x] **3. Add a real "unknown" state.**
Everything that isn't a confirmed takeover currently lands in `safe` — including unknown services and failed probes. Fix: use three states — `vulnerable` / `not_vulnerable` / `unknown` — and make sure DNS errors, HTTP failures, and unsupported services all map to `unknown`, never `safe`/`not_vulnerable`.

## C. Fingerprint accuracy

- [x] **4. S3 fingerprint is too broad.** Matches any `amazonaws.com` CNAME (API Gateway, ELB, etc.), not just S3. Fix: match actual S3 endpoint patterns (regex), and add a regression test for a non-S3 `amazonaws.com` host that should *not* match.

- [ ] **5. Fastly fingerprint is too weak.** `fastly.net` in the CNAME + generic "Fastly error" body isn't enough — legit configs error too. Fix: combine CNAME pattern + HTTP status + specific body marker + relevant headers.

- [ ] **6. Replace substring CNAME matching with patterns.** `if fingerprint["cname_contains"] in cname` is broad and won't scale. Fix: let fingerprints define regex patterns; normalize case and trailing dot before matching.

[...]

PR #159, opened 2026-08-13
fix(subdomain-takeover): anchor the AWS S3 fingerprint on documented bucket endpoints

Nitjsefnie:
## Closes

Partially addresses #154 (item 4)

## Summary

The AWS S3 takeover fingerprint matched any CNAME containing `amazonaws.com`, so API Gateway, ELB and other AWS services were fingerprinted as S3. It now matches only the documented S3 *bucket* endpoint hosts — the ones a dangling CNAME can actually leave answering `NoSuchBucket`.

## What changed

- `tools/subdomain_takeover_tool.py`: the S3 entry's `cname_contains: "amazonaws.com"` becomes a `cname_pattern` regex covering the documented bucket endpoint families — under `amazonaws.com`: virtual-hosted regional, legacy dash, legacy global, dual-stack, FIPS (± dual-stack), Transfer Acceleration (± dual-stack), and both website forms; under `amazonaws.com.cn`: regional, legacy dash, legacy global, dual-stack, and website. Sources are linked in the code comment.
- Regions are matched by shape (`us-east-1`, `us-gov-west-1`, `cn-north-1`, `us-iso-east-1`) rather than an enumeration, so the fingerprint does not go stale as AWS adds regions.
- The pattern is anchored on a label boundary and on the end of the host, so `s3` inside a longer label (`nots3.amazonaws.com`) and lookalike suffixes (`...s3.us-east-1.amazonaws.com.evil.com`) do not match.
- `_match_fingerprint` accepts `cname_pattern` alongside the existing `cname_contains` and strips the trailing dot before matching. Every other fingerprint is untouched, and the resolved CNAME reported in the result keeps its original casing.
- `tests/test_subdomain_takeover.py`: one test asserting selection over 18 hosts that must match and 18 that must not — non-S3 AWS services (API Gateway, ELB), the sibling `s3-*` endpoints that are not buckets (`s3-control`, `s3-accesspoint`, `s3-object-lambda`, `s3-outposts`, `s3express-*`), fabricated region tokens, endpoint forms AWS does not document for a given partition, and the label/suffix evasions above.

## Notes

- The sibling `s3-*` endpoints are excluded deliberately: they are not buckets and never answer `NoSuchBucket`, so matching them would reproduce the same false positive in a narrower shape.
- The China partition documents fewer forms than the commercial one (no FIPS, no Transfer Acceleration, website is dot-separated only). Those are rejected rather than accepted for symmetry, and the tests pin that.
- Shape-matched regions mean a syntactically valid but nonexistent region still matches; the narrowing here comes from anchoring on endpoint families, not from validating region names.
- Scope is item 4 only — the other items in #154 are untouched.

## Verification

- [x] All tests pass (`pytest tests/ -v`)
- [x] Updated Documentation ( if required ) — none required

The full suite ran on a GitHub runner against this branch head (run `31689494665`), whose log records the checked-out SHA as `409a8c6d6caa3c8aa2891ca4edbd77e76d1d0961`: **313 passed, 15 subtests passed**. The new node `tests/test_subdomain_takeover.py::test_s3_fingerprint_matches_only_s3_endpoints` passes, driving all 36 hosts through the tool's real resolution path.

Generated by GPT-5.6 Sol (brief, review), Kimi K3 (implementation, verification), Claude Opus 5 (implementation, verification, testing, fork CI)

### DIFF
--- a/tests/test_subdomain_takeover.py
+++ b/tests/test_subdomain_takeover.py
@@ -246,6 +246,53 @@ def test_both_schemes_fail_is_safe(mock_enum, mock_resolver_class, mock_get):
     assert mock_get.call_count == 2
 
 
+@patch("tools.subdomain_takeover_tool.requests.get")
+@patch("tools.subdomain_takeover_tool.dns.resolver.Resolver")
+@patch("tools.subdomain_takeover_tool.dns_enumeration")
+def test_s3_fingerprint_matches_only_s3_endpoints(mock_enum, mock_resolver_class, mock_get):
+    """Only actual S3 endpoint CNAMEs select the AWS S3 fingerprint; other amazonaws.com services are safe and unprobed."""
+    mock_enum.return_value = _enumeration_result(["static.example.com"])
+    resolver = Mock()
+    mock_resolver_class.return_value = resolver
+
+    s3_cnames = [
+        "static-example-com.s3.amazonaws.com.",
+        "static-example-com.s3.us-east-1.amazonaws.com.",
+        "static-example-com.s3-us-west-2.amazonaws.com.",
+        "static-example-com.s3.dualstack.us-east-1.amazonaws.com.",
+        "static-example-com.s3-website-us-east-1.amazonaws.com.",
+        "static-example-com.s3-website.eu-west-1.amazonaws.com.",
+        "Static-Example-Com.S3.Us-East-1.Amazonaws.Com.",
+    ]
+    non_s3_cnames = [
+        "abc123def4.execute-api.us-east-1.amazonaws.com.",  # API Gateway
+        "my-alb-1234567890.us-east-1.elb.amazonaws.com.",  # Elastic Load Balancing
+        "dualstack.my-alb-1234567890.us-west-2.elb.amazonaws.com.",  # ELB dualstack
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
# Review: Approved

Great work on this PR. The S3 fingerprinting is now much more precise instead of treating every `amazonaws.com` CNAME as an S3 endpoint.

I especially liked the regression coverage here — you covered multiple valid S3 endpoint formats as well as non-S3 AWS services like API Gateway, ELB, S3 Control, Access Points, Object Lambda, Outposts, and S3 Express. The assertions that non-S3 endpoints are not even HTTP-probed are also a nice touch. Everything passes on my side. 👍

**Non-blocking suggestion:** The inline comments around the S3 endpoint regex are quite detailed, and some of the endpoint-specific behavior is already well documented by the regression tests. You could consider shortening those comments to focus on the matching strategy/rationale. If you agree, feel free to clean this up in a follow-up PR — definitely not a blocker for this one.

Going to merge it

Nitjsefnie:
Agreed — the endpoint-by-endpoint notes duplicate what the regression tests already pin, so the comments should be cut back to the matching strategy and why the pattern is anchored on the documented bucket endpoint forms. I will send that as a follow-up PR once this one lands.

Issue #144, opened 2026-08-04
Enhancement: Improve DNS Enumeration Coverage, Performance, and Error Handling


gaoharimran29-glitch:
## Summary

The current `dns_enumeration` implementation is well-structured and provides support for many commonly used DNS record types. However, there are several opportunities to improve its reliability, performance, and the amount of information it can return.

These enhancements are independent of one another and **do not need to be completed in a single pull request**. Contributors are welcome to work on **one or more items in separate PRs**.

### 1. Replace broad exception handling with specific DNS exceptions

The current implementation catches `Exception` in multiple places. While this prevents crashes, it also hides the actual cause of lookup failures and can make debugging more difficult.

Instead, catch specific exceptions such as:

* `dns.resolver.NoAnswer`
* `dns.resolver.NXDOMAIN`
* `dns.resolver.Timeout`
* `dns.resolver.NoNameservers`
* Other relevant `dnspython` exceptions where appropriate.

This makes failures easier to understand and avoids masking unexpected programming errors.

---

### 2. Perform DNS lookups concurrently

DNS lookups are network-bound operations. The current implementation performs every query sequentially, which increases the total execution time, especially when querying multiple record types and brute-forcing subdomains.

Using `concurrent.futures.ThreadPoolExecutor` would allow multiple independent DNS requests to run in parallel, significantly improving overall performance without changing the public API.

---

### 3. Prefer UDP first, then fall back to TCP

Currently every DNS query is performed using:

```python
resolver.resolve(..., tcp=True)
```

In normal DNS resolution, clients typically use **UDP first** and retry with **TCP only when necessary** (for example, if the response is truncated).

Following this behavior would reduce latency while still maintaining compatibility with larger DNS responses.

---

### 4. Add DNSSEC-related record support

To improve security-focused reconnaissance, consider enumerating DNSSEC records such as:

* `DNSKEY`
* `DS`
* `RRSIG`
* `NSEC` / `NSEC3` (where applicable)

These records provide valuable information about DNSSEC deployment and domain security.

---

### 5. Expand subdomain discovery

The current brute-force list contains only a small set of common subdomains.

Expanding the built-in wordlist (or supporting configurable/custom wordlists) would significantly improve discovery coverage while keeping the implementation simple. We have to extend it to atleast 50 for better subdomain finding.

---

### 6. Enumerate SRV records

Many enterprise services publish DNS SRV records.

[...]

PR #160, opened 2026-08-13
fix(dns): catch specific dnspython lookup errors in dns_enumeration (#144 item 1)

Nitjsefnie:
## Closes

Partially addresses #144 (item 1)

## Summary

Replaces the broad `except Exception` handlers in `dns_enumeration` with the specific lookup failures dnspython documents for `Resolver.resolve()`. Ordinary negative DNS answers are still recorded as empty/missing, while unexpected programming errors now propagate as defects instead of being silently swallowed.

## What changed

- `tools/dns_tool.py`: added a `LOOKUP_ERRORS` tuple (`NoAnswer`, `NoNameservers`, `YXDOMAIN`, `Timeout`) and a `SUBDOMAIN_LOOKUP_ERRORS` tuple (adds `NXDOMAIN` and `NameTooLong`, the ordinary "no such subdomain" outcomes on the brute-force path). The record loop now catches only `LOOKUP_ERRORS` (target-domain `NXDOMAIN` still ends the enumeration as "domain does not exist"), and the subdomain loop continues only on `SUBDOMAIN_LOOKUP_ERRORS`.
- `tests/test_dns_enumeration.py`: focused regression tests pinning each distinct resolver outcome, including that an unexpected error (e.g. `RuntimeError`) propagates from both the record loop and the subdomain brute-force loop.

## Notes

- Scoped to item 1 of #144 only; the remaining enhancement items are intentionally left out and stay open for separate PRs.
- Anything outside the documented lookup-error set is treated as a defect and left to propagate, rather than reported as an empty result.

## Verification

- [x] Ran locally and confirmed it works as expected (focused `tests/test_dns_enumeration.py` runs)
- [x] All tests pass (`pytest tests/ -v` → 316 passed, 25 subtests passed; fork CI run https://github.com/Nitjsefnie-OSC/AynOps/actions/runs/31691285314, log-confirmed checkout of edb80ad6bda39b7be7888084a5099af31cfc9f03)
- [x] Updated Documentation (not required — internal error-handling change, no docs impact)

Generated by GPT-5.6 Sol (brief, review), Claude Opus 5 (implementation), Kimi K3 (implementation, verification, testing, fork CI)

### DIFF
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -1,3 +1,5 @@
+import dns.exception
+import dns.name
 import dns.resolver
 from utils.helpers import is_valid_domain, normalize_domain
 
@@ -24,6 +26,25 @@ COMMON_SUBDOMAINS = [
 
 # A subdomain counts as found if any of these record types resolves for it.
 SUBDOMAIN_RECORD_TYPES = ("A", "AAAA", "CNAME")
+
+# Lookup failures dnspython documents for Resolver.resolve(): the name has no
+# RRset of the requested type, no nameserver could answer, the resolution
+# lifetime expired, or the name is too long after DNAME substitution. NXDOMAIN
+# is documented too but handled separately, because a target domain that does
+# not exist ends the whole enumeration. Anything outside this set is a defect
+# rather than a DNS answer, and is left to propagate.
+LOOKUP_ERRORS = (
+    dns.resolver.NoAnswer,
+    dns.resolver.NoNameservers,
+    dns.resolver.YXDOMAIN,
+    # dns.resolver.Timeout is dns.resolver.LifetimeTimeout, a subclass of this.
+    dns.exception.Timeout,
+)
+
+# On the brute-force path NXDOMAIN is the ordinary result for a name that does
+# not exist, and a candidate built by prefixing a label to a long target can
+# exceed the 255-octet limit; both mean "no such subdomain", not a failed scan.
+SUBDOMAIN_LOOKUP_ERRORS = LOOKUP_ERRORS + (dns.resolver.NXDOMAIN, dns.name.NameTooLong)
 
 
 def _clean_name(value) -> str:
@@ -119,11 +140,9 @@ def dns_enumeration(domain: str) -> dict:
                 records[rtype] = [_clean_name(r) for r in answers]
             else:
                 records[rtype] = [str(r) for r in answers]
-        except dns.resolver.NoAnswer:
-            records[rtype] = []
         except dns.resolver.NXDOMAIN:
             return {"success": False, "error": f"Domain {domain} does not exist"}
-        except Exception:
+        except LOOKUP_ERRORS:
             records[rtype] = []
 
     # Subdomain brute-force (common subdomains); a subdomain counts as found
@@ -138,7 +157,7 @@ def dns_enumeration(domain: str) -> dict:
                 resolver.resolve(full, rtype, lifetime=SUBDOMAIN_LIFETIME, tcp=True)
                 found_subdomains.append(full)
                 break
-            except Exception:
+            except SUBDOMAIN_LOOKUP_ERRORS:
                 continue
 
     return {
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -31,8 +31,8 @@ SUBDOMAIN_RECORD_TYPES = ("A", "AAAA", "CNAME")
 # RRset of the requested type, no nameserver could answer, the resolution
 # lifetime expired, or the name is too long after DNAME substitution. NXDOMAIN
 # is documented too but handled separately, because a target domain that does
-# not exist ends the whole enumeration. Anything outside this set is a defect
-# rather than a DNS answer, and is left to propagate.
+# not exist ends the whole enumeration. Anything outside this set is recorded
+# as an unexpected lookup error so the rest of the scan can continue.
 LOOKUP_ERRORS = (
     dns.resolver.NoAnswer,
     dns.resolver.NoNameservers,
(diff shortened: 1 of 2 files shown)


tools/dns_tool.py
```diff
@@ -119,11 +140,9 @@ def dns_enumeration(domain: str) -> dict:
                 records[rtype] = [_clean_name(r) for r in answers]
             else:
                 records[rtype] = [str(r) for r in answers]
-        except dns.resolver.NoAnswer:
-            records[rtype] = []
         except dns.resolver.NXDOMAIN:
             return {"success": False, "error": f"Domain {domain} does not exist"}
-        except Exception:
+        except LOOKUP_ERRORS:
             records[rtype] = []
```
gaoharimran29-glitch: The issue's description stated reason for the change was: *"it also hides the actual cause of lookup failures and can make debugging more difficult."* Right now `NoAnswer`, `Timeout`, and `NoNameservers` all still collapse to the same `records[rtype] = []` — the cause is still hidden. Please add per-record-type error capture and return errors alongside records.

gaoharimran29-glitch (changes requested):
## Review: Changes Requested

The exception narrowing looks solid. Replacing the broad `except Exception` with the appropriate `dnspython` exceptions is a good improvement, and the separation of `LOOKUP_ERRORS` and `SUBDOMAIN_LOOKUP_ERRORS` correctly ensures that `NXDOMAIN` / `NameTooLong` are treated as "not found" only for subdomain lookups, not for the target domain.

There are two things to address before merge:

### 1. Surface the actual cause of expected DNS failures (original issue) 
Check the inline comment

### 2. Unexpected exceptions should not crash the entire tool (new ask, beyond the original issue)

Separately, while reviewing this I'd like unexpected exceptions to also not crash the whole tool. Please catch these as well, but record them distinctly from expected DNS errors so they remain identifiable, e.g.:

```python
errors[rtype] = f"unexpected: {type(exc).__name__}"
```

The goal is for the tool to **never fully fail** — it should always return whatever it was able to gather, with clear error information for anything it couldn't retrieve.

Because of this, the following tests will need to be rewritten to assert graceful handling instead of `assertRaises`:

- `test_unexpected_error_propagates_from_record_lookup`
- `test_unexpected_error_propagates_from_subdomain_lookup`

### Tests

The existing tests provide good coverage for the exception narrowing and (current) propagation behavior. Once the `errors` field is added for both expected and unexpected failures, please add explicit assertions for the recorded error values (e.g. `result["errors"]["TXT"] == "NoAnswer"`) so this behavior is protected going forward.

---

Overall, the direction here is good — these changes should make the tool more resilient while keeping failures visible and understandable.

gaoharimran29-glitch (approved):
Thanks for addressing the review comments! Everything looks good now, including the improved exception handling and test coverage. Really appreciate the effort you put into making the implementation more robust.

Great contribution overall — hope to see more contributions from you in the future!

Nitjsefnie:
Thanks for the approval — but please hold the merge briefly, because my own post-review check found a problem in something I added that you did not ask for.

The `subdomain_errors` map is mine, not part of your ask: your inline comment was on the record loop. On a realistic scan it produces 190 entries that are all `NXDOMAIN`, about 5.1KB of a 5.8KB payload, and it restates for each candidate exactly what its absence from `subdomains_found` already says. It also contradicts the comment in the module itself — that NXDOMAIN there means "no such subdomain", not a failed scan — and so contradicts the `LOOKUP_ERRORS` / `SUBDOMAIN_LOOKUP_ERRORS` split you called out as correct. For a tool whose output lands in a model context, that is a lot of bytes saying nothing.

Worse, it can contradict itself: a host that fails on `A` and resolves on `AAAA` ends up listed in `subdomains_found` *and* carrying an error for the same name.

Separately, my `except Exception` ended up spanning the answer-parsing block rather than just the resolver call, so a bug in my own formatting code would surface as `unexpected: AttributeError` in the result with `success: true`, instead of failing loudly. That is the opposite of the visibility you asked for.

Correction on the way shortly: expected subdomain failures stop being recorded (unexpected ones still are, per your ask, and still do not crash the scan), and the catch narrows to the resolver call. `result["errors"]["TXT"] == "NoAnswer"`, the NXDOMAIN-on-target behaviour and the existing keys are all unchanged. I will re-request review once it is pushed and green — sorry for the extra round trip.

Nitjsefnie:
Our messages crossed — you merged about twenty seconds before my note landed, so please ignore the "hold the merge" framing. Nothing is broken in what shipped; the three points stand as follow-up work rather than as a blocker, and I will send them as a separate PR against `main` shortly:

1. drop the `subdomain_errors` map for *expected* failures (it was mine, not part of your ask, and on a realistic scan it is ~5.1KB of a 5.8KB payload restating what absence from `subdomains_found` already says — unexpected failures stay recorded, as you asked);
2. stop a host that fails `A` and resolves `AAAA` from appearing as both found and errored;
3. narrow my `except Exception` to the resolver call, so a bug in the answer-parsing code fails loudly instead of arriving as `unexpected: AttributeError` with `success: true`.

`result["errors"]["TXT"] == "NoAnswer"`, the NXDOMAIN-on-target behaviour and the existing keys are untouched by all three.

Thanks for the review and the quick turnaround.

gaoharimran29-glitch:
@Nitjsefnie No worries at all — and honestly, great catch. I appreciate you doing the post-review check and spotting these issues yourself.

The `subdomain_errors` payload bloat and the overly broad `except Exception` are both valid concerns, especially the latter since we don't want genuine implementation bugs to get silently reported as scan errors.

Since the PR is already merged, please go ahead with the correction as a follow-up PR.

Thanks for being proactive about this and for taking ownership of the implementation. Looking forward to the follow-up PR — and definitely keep contributing!

PR #162, opened 2026-08-14
fix(dns): cut expected-failure noise from the payload and contain remote decode errors

Nitjsefnie:
## Closes

Follow-up to #160 (merged), at your request in https://github.com/AynOps/AynOps/pull/160#issuecomment-5290384585 — the three points I raised against my own merged work.

## Summary

Cuts the DNS enumeration payload from ~5.8KB to ~0.8KB on a typical scan by no longer recording expected subdomain failures, stops a host being reported as both found and errored, and narrows the exception handling so a bug in this repo's own formatting code fails loudly instead of arriving as a scan result.

## What changed

- **Expected subdomain failures are no longer recorded.** They were 190 all-`NXDOMAIN` entries on one realistic scan — about 5.1KB of a 5.8KB payload — restating for each candidate exactly what its absence from `subdomains_found` already says, and contradicting the comment in `dns_tool.py` that NXDOMAIN there means "no such subdomain, not a failed scan". **Unexpected** subdomain failures are still recorded, as you asked.
- **A host that fails `A` and resolves `AAAA` no longer appears in both `subdomains_found` and the error map.** Closed for every member of `SUBDOMAIN_LOOKUP_ERRORS`.
- **`except Exception` narrowed to the resolver call** rather than the whole answer-parsing block. A `TypeError` or `IndexError` in this repo's formatting code now propagates instead of surfacing as `unexpected: AttributeError` with `success: true`. As a side effect `ttl` and `errors` are now mutually exclusive by construction, which also fixes a pre-existing case where a record type could be reported as failed while still carrying a TTL.
- Untrusted record bytes are contained. TXT rdata and CAA values are arbitrary octet strings under RFC 1035, and narrowing the `try` had exposed the `.decode("utf-8")` calls — one non-UTF-8 record would have escaped `dns_enumeration()` and, through the caller in `subdomain_takeover_tool.py`, taken that tool's whole output with it. That is contained and reported now, without re-swallowing genuine bugs.

## Notes

Two things left deliberately, both matching current behaviour rather than changing it:

- One malformed TXT string discards the whole TXT rrset rather than only the offending record.
- The recorded label for an undecodable record reads `unexpected: UnicodeDecodeError`, which now names a condition the code explicitly anticipates. Renaming it would change the output shape you just reviewed, so I left it — happy to change it if you'd rather.

The empty `subdomain_errors` key is retained even when there is nothing to report, so the unexpected-error path keeps a stable shape. It is 20 bytes; say the word and I'll drop it entirely when empty.

## Verification

- [x] All tests pass (`pytest tests/` — 320 passed, 1 skipped, 25 subtests; the skip is pre-existing, `cryptography` is not in `requirements.txt`)
- [x] Updated Documentation ( if required ) — not required, no user-facing docs describe this shape

Each new assertion was checked by reverting the production change and confirming it fails by `AssertionError` rather than passing vacuously. The decode containment was exercised against real `dns.rdata.from_wire` records rather than mocks — 26,000 cases across every formatter branch, none escaping — while injected `TypeError`, `IndexError`, `AttributeError`, `ValueError` and `KeyError` all still propagate. The DNS tests were also re-run with network syscalls patched to raise, to confirm nothing reaches the network.

---

Generated by GPT-5.6-Luna (implementation), Claude Opus 5 (brief, review, testing)

### DIFF
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -127,30 +127,38 @@ def dns_enumeration(domain: str) -> dict:
             ttl = _record_ttl(answers)
             if ttl is not None:
                 ttls[rtype] = ttl
-            if rtype == "MX":
-                records[rtype] = [
-                    {"preference": r.preference, "exchange": _clean_name(r.exchange)}
-                    for r in answers
-                ]
-            elif rtype == "SOA":
-                r = answers[0]
-                records[rtype] = {
-                    "mname": _clean_name(r.mname),
-                    "rname": _clean_name(r.rname),
-                    "serial": r.serial,
-                    "refresh": r.refresh,
-                    "retry": r.retry,
-                    "expire": r.expire,
-                    "minimum": r.minimum
-                }
-            elif rtype == "TXT":
-                records[rtype] = [_format_txt_record(r) for r in answers]
-            elif rtype == "CAA":
-                records[rtype] = [_format_caa_record(r) for r in answers]
-            elif rtype in {"NS", "CNAME"}:
-                records[rtype] = [_clean_name(r) for r in answers]
-            else:
-                records[rtype] = [str(r) for r in answers]
+            try:
+                if rtype == "MX":
+                    records[rtype] = [
+                        {"preference": r.preference, "exchange": _clean_name(r.exchange)}
+                        for r in answers
+                    ]
+                elif rtype == "SOA":
+                    r = answers[0]
+                    records[rtype] = {
+                        "mname": _clean_name(r.mname),
+                        "rname": _clean_name(r.rname),
+                        "serial": r.serial,
+                        "refresh": r.refresh,
+                        "retry": r.retry,
+                        "expire": r.expire,
+                        "minimum": r.minimum
+                    }
+                elif rtype == "TXT":
+                    records[rtype] = [_format_txt_record(r) for r in answers]
+                elif rtype == "CAA":
+                    records[rtype] = [_format_caa_record(r) for r in answers]
+                elif rtype in {"NS", "CNAME"}:
+                    records[rtype] = [_clean_name(r) for r in answers]
+                else:
+                    records[rtype] = [str(r) for r in answers]
+            except UnicodeDecodeError as exc:
+                # TXT chunks and CAA values are arbitrary remote octets, not
+                # guaranteed UTF-8. Keep this record type's failure visible
+                # without swallowing defects in our own formatting logic.
+                records[rtype] = []
+                errors[rtype] = f"unexpected: {type(exc).__name__}"
+                ttls.pop(rtype, None)
 
     # Subdomain brute-force (common subdomains); a subdomain counts as found
     # when any of A/AAAA/CNAME resolves, so IPv6-only and aliased hosts are
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -157,7 +157,7 @@ def dns_enumeration(domain: str) -> dict:
                 # guaranteed UTF-8. Keep this record type's failure visible
                 # without swallowing defects in our own formatting logic.
                 records[rtype] = []
-                errors[rtype] = f"unexpected: {type(exc).__name__}"
+                errors[rtype] = type(exc).__name__
                 ttls.pop(rtype, None)
 
     # Subdomain brute-force (common subdomains); a subdomain counts as found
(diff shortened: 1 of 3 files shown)


tools/dns_tool.py
```diff
@@ -163,8 +173,7 @@ def dns_enumeration(domain: str) -> dict:
                 resolver.resolve(full, rtype, lifetime=SUBDOMAIN_LIFETIME, tcp=True)
                 found_subdomains.append(full)
                 break
-            except SUBDOMAIN_LOOKUP_ERRORS as exc:
-                subdomain_errors.setdefault(full, {})[rtype] = type(exc).__name__
+            except SUBDOMAIN_LOOKUP_ERRORS:
                 continue
             except Exception as exc:
```
gaoharimran29-glitch, line 181: I tested the unexpected-error path and found that the current implementation can still report the same hostname in both `subdomains_found` and `subdomain_errors`.

For example:

```text
A     -> unexpected exception
AAAA  -> resolves successfully
```

The `A` failure is persisted in `subdomain_errors`, and when `AAAA` subsequently succeeds the hostname is added to `subdomains_found`, but the previously recorded error is not cleared.

The existing `test_aaaa_found_subdomain_has_no_expected_a_error` does not exercise this path because `NoAnswer` is covered by `SUBDOMAIN_LOOKUP_ERRORS` and therefore never enters the error map.

The PR description says that a host failing one record type and resolving another should not appear in both collections, although the qualification "for every member of `SUBDOMAIN_LOOKUP_ERRORS`" could mean that this invariant is intentionally limited to expected lookup failures.

Could you clarify the intended invariant here? If unexpected lookup errors should also be cleared once another record type establishes the host, please add a regression test for this path and adjust the implementation accordingly.

tools/dns_tool.py
```diff
@@ -149,6 +123,42 @@ def dns_enumeration(domain: str) -> dict:
...
+                elif rtype in {"NS", "CNAME"}:
+                    records[rtype] = [_clean_name(r) for r in answers]
+                else:
+                    records[rtype] = [str(r) for r in answers]
+            except UnicodeDecodeError as exc:
+                # TXT chunks and CAA values are arbitrary remote octets, not
+                # guaranteed UTF-8. Keep this record type's failure visible
+                # without swallowing defects in our own formatting logic.
+                records[rtype] = []
+                errors[rtype] = f"unexpected: {type(exc).__name__}"
```
gaoharimran29-glitch: Non-blocking: `UnicodeDecodeError` is explicitly anticipated and handled in this branch, so classifying it as `unexpected` is semantically misleading.

Please use a more specific classification for this case (e.g. a decode/data-format error) so it remains distinguishable from genuinely unexpected exceptions. The handling itself looks correct; this is purely an error-classification/wording change.

gaoharimran29-glitch (changes requested):
Overall, this is a strong follow-up and a substantial improvement over the previous implementation. The tighter exception boundaries, explicit handling of untrusted DNS record data, and the depth of the regression coverage all look well thought out.

I’ve left one blocking comment around the subdomain error/found-state invariant and one non-blocking comment on error classification. Please see the inline comments for the specific details.

Really good work on the testing here as well, especially the effort to verify that genuine formatting/programming exceptions still propagate instead of being silently converted into scan results. The scope is focused and the changes are generally well structured.

gaoharimran29-glitch:
Also, a seperate thing from review: You've been contributing consistently to AynOps and the quality of your work has really stood out. I'd like to invite you to join the AynOps organization as a member if you're interested. Would you be up for that?

Nitjsefnie:
Thank you — that means a lot, but I'd rather stay an outside contributor for now. I'll keep sending PRs and working through review exactly the same way, and I'm glad the work has been useful.

gaoharimran29-glitch:
> Thank you — that means a lot, but I'd rather stay an outside contributor for now. I'll keep sending PRs and working through review exactly the same way, and I'm glad the work has been useful.
> 

No worries at all! Completely understand. Really appreciate your contributions to AynOps — glad to have you around!

Nitjsefnie:
Both are in `ec18571`.

On the invariant, the stronger reading is the intended one: a host that resolves on any record type should never appear in `subdomain_errors`, whatever kind of exception an earlier record type raised. A successful resolve now clears any entry already recorded for that host, so the two collections cannot disagree. The new regression test drives an exception that is deliberately *not* a member of `SUBDOMAIN_LOOKUP_ERRORS` — which is precisely why the existing test could not reach this path — and it fails on its assertion, not an error, without the fix.

For the classification: the decode case now records the bare exception name, matching what the file already does for anticipated `LOOKUP_ERRORS`, so the `unexpected:` prefix stays reserved for genuine surprises.

gaoharimran29-glitch (approved):
Approved. The follow-up changes look solid and the implementation now closes the edge case around stale subdomain errors when a later record type successfully resolves the host.

The additional regression coverage validates the unexpected-error path properly, and the error classification cleanup also makes the output semantics clearer. The exception-boundary changes, untrusted DNS record handling, and overall test coverage are all well considered.

Thanks for the thorough follow-up and for addressing the review feedback cleanly.

PR #163, opened 2026-08-14
feat(dns): enumerate SRV records for common services

Nitjsefnie:
## Closes

Addresses item 6 of #144 (SRV records). Items 4, 7 and 8 are deliberately left for separate PRs.

## Summary

`dns_enumeration` now probes the five SRV services listed in item 6 and reports them under `srv_records` / `srv_errors`, keeping a service with no SRV record distinct from a service whose lookup failed.

## What changed

- `tools/dns_tool.py`: new `SRV_SERVICES` constant with the five owner-name prefixes item 6 names — `_sip._tcp`, `_ldap._tcp`, `_xmpp-client._tcp`, `_kerberos._tcp`, `_autodiscover._tcp`.
- `tools/dns_tool.py`: SRV enumeration loop using the module's existing resolver call shape; answers parsed to `{priority, weight, port, target}` with the target passed through `_clean_name`.
- `tools/dns_tool.py`: result dict gains `srv_records` and `srv_errors`.
- `tests/test_dns_enumeration.py`: coverage for a successful lookup, a service with no record, and a service whose lookup raises — all against mocked resolvers.

## Notes

Anticipated SRV lookup failures — no nameserver, YXDOMAIN, lifetime timeout, and an owner name pushed past 255 octets by the service prefix — are recorded by their bare exception name in `srv_errors`, the same convention the record-type loop uses, so the `unexpected:` prefix keeps its meaning of "a genuine surprise worth investigating".

That last case is worth calling out: prefixing `_autodiscover._tcp.` to a long-but-valid domain can push the owner name past 255 octets, so `NameTooLong` is reachable for a domain this tool's own `is_valid_domain` accepts. It is anticipated, not a surprise, and the SRV loop now treats the same six exceptions as anticipated that the subdomain loop already did — the two paths no longer classify identical input differently.

A no-record service and a failed lookup stay distinguishable: the first appears in neither map, the second only in `srv_errors`. Collapsing them fails a test.

## Verification

- [x] All tests pass (`python -m pytest tests/test_dns_enumeration.py`) — full suite green on a runner: `324 passed, 32 subtests passed`
- [x] Updated Documentation ( if required ) — no doc change needed; no user-facing docs describe the DNS result shape

Generated by Kimi K3 (implementation, testing), Claude Opus 5 (brief, review)

### DIFF
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -57,6 +57,11 @@ LOOKUP_ERRORS = (
 # exceed the 255-octet limit; both mean "no such subdomain", not a failed scan.
 SUBDOMAIN_LOOKUP_ERRORS = LOOKUP_ERRORS + (dns.resolver.NXDOMAIN, dns.name.NameTooLong)
 
+# An SRV owner name is built by prefixing a service label to the target, so a
+# target near the length limit pushes the queried name past 255 octets; like
+# the brute-force path, that means "no such service name", not a failed scan.
+SRV_LOOKUP_ERRORS = LOOKUP_ERRORS + (dns.name.NameTooLong,)
+
 
 def _clean_name(value) -> str:
     return str(value).rstrip(".")
@@ -187,7 +192,7 @@ def dns_enumeration(domain: str) -> dict:
             answers = resolver.resolve(srv_name, "SRV", lifetime=RESOLVER_LIFETIME, tcp=True)
         except (dns.resolver.NXDOMAIN, dns.resolver.NoAnswer):
             srv_records[service] = []
-        except LOOKUP_ERRORS as exc:
+        except SRV_LOOKUP_ERRORS as exc:
             srv_records[service] = []
             srv_errors[service] = type(exc).__name__
         except Exception as exc:
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -27,6 +27,17 @@ COMMON_SUBDOMAINS = [
 # A subdomain counts as found if any of these record types resolves for it.
 SUBDOMAIN_RECORD_TYPES = ("A", "AAAA", "CNAME")
 
+# SRV services probed for the target domain (issue #144, item 6: SIP, LDAP,
+# XMPP, Kerberos and Autodiscover), as the owner-name prefixes queried with
+# the SRV record type.
+SRV_SERVICES = [
+    "_sip._tcp",
+    "_ldap._tcp",
+    "_xmpp-client._tcp",
+    "_kerberos._tcp",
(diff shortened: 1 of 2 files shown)


tools/dns_tool.py
```diff
@@ -27,6 +27,17 @@
...
+# XMPP, Kerberos and Autodiscover), as the owner-name prefixes queried with
+# the SRV record type.
+SRV_SERVICES = [
+    "_sip._tcp",
+    "_ldap._tcp",
+    "_xmpp-client._tcp",
+    "_kerberos._tcp",
+    "_autodiscover._tcp",
+]
+
```
gaoharimran29-glitch, line 40: Broader SRV coverage:
A possible follow-up improvement would be to broaden SRV coverage to additional common service/protocol variants, for example `_sip._udp`,` _ldap._udp`,` _xmpp-server._tcp`,` _kerberos._udp`,` _kerberos-master._tcp`,` _kerberos-master._udp`, `_kpasswd._tcp`, and `_kpasswd._udp`.

This is not a blocker for this PR, since these additional variants are not explicitly specified in the issue description and the current implementation satisfies the requested scope. If desired, the broader coverage can be addressed in a follow-up PR.

tools/dns_tool.py
```diff
@@ -27,6 +27,17 @@
 # A subdomain counts as found if any of these record types resolves for it.
 SUBDOMAIN_RECORD_TYPES = ("A", "AAAA", "CNAME")
 
+# SRV services probed for the target domain (issue #144, item 6: SIP, LDAP,
+# XMPP, Kerberos and Autodiscover), as the owner-name prefixes queried with
+# the SRV record type.
```
gaoharimran29-glitch, line 32: Keep code comments implementation-focused:
One small cleanup suggestion is to avoid comments that reference the issue number or specific issue items, such as issue #144, item 6. Those references are useful in the PR/issue discussion but can become stale or add unnecessary noise once the code is merged. It would be cleaner for the comment to simply describe what the constant represents and why those SRV owner-name prefixes are being queried.

gaoharimran29-glitch (approved):
## Review: Approved

Thanks for the solid implementation. The SRV enumeration covers all services called out in the issue (SIP, LDAP, XMPP, Kerberos, and Autodiscover), correctly parses the SRV priority/weight/port/target fields, handles missing records and resolver errors appropriately, and includes good test coverage for successful lookups and anticipated failure cases.

The implementation looks correct and is ready to merge. **Approved.**

However, there are two non blocking suggestion mentioned in the inline comment.
Both suggestions are non-blocking and can be handled in a follow-up PR if desired.

PR #171, opened 2026-08-20
refactor(techstack): extract the technology fingerprinting engine

Nitjsefnie:
## Closes

Nothing on its own. This is groundwork for issue 130 rather than its completion — Phase 1 landed in #141, and this adds the engine seam Phases 2–4 need without changing any detection behaviour. Refs #130

## Summary

Extracts the technology-fingerprinting logic out of `tech_stack_detect` into a dedicated `tools/fingerprint/` engine, with the signature tables separated from the matching code. Output is unchanged for every response a real `requests` session can produce; the value here is structural, so the detection work in later phases has somewhere to land.

## What changed

- New `tools/fingerprint/signatures.py` holds the four signature tables — 27 technologies, 53 markers — as data, carried over unmodified from their inline definitions.
- New `tools/fingerprint/engine.py` exposes `fingerprint(raw_headers, raw_text)`, split into a `headers_layer` and an `html_layer`. Those are two of the layers issue 130 names under Phase 2, now separable, so cookie, meta-generator and asset layers can be added without reopening the tool module.
- `tools/techstack_tool.py` drops to fetch, delegate, assemble. `tech_stack_detect(domain) -> dict` and its output shape are untouched, so `server.py` and `tools/signals/registry.py` need no adjustment.
- `tests/test_tech_stack_detect.py` gains characterisation tests written **before** the move: all 53 markers individually, technology key order, within-category table order, case handling, the preserved false positives, and the error branches.
- `tests/test_fingerprint_engine.py` covers the new seam directly, including layer independence and table integrity.
- README gains a short note on the new module.

## Notes

- **Byte-for-byte equivalence was the acceptance bar, and three of the five commits exist only to hold it.** The caller now drains and normalises the response headers through `.items()` *before* `resp.text` is read, so the engine receives exactly what the old inline code computed. That ordering is load-bearing in a non-obvious way: reading the body can drain a streamed response, and the original code interleaved the header and body reads.
- **Known false positives are preserved rather than quietly repaired.** The Google Analytics marker is still the bare string `ga(`, so a page containing `saga(` still matches. Issue 130 lists false positives among the problems to solve, but fixing them inside a behaviour-preserving refactor would bury the change in a diff that claims to change nothing — they belong to the phase that addresses detection quality.
- **One redundancy I would rather flag than hide:** the caller normalises header keys and `headers_layer` normalises them again. For anything a real `requests.Session` returns this is a no-op, because headers arrive as a `CaseInsensitiveDict` of plain strings. It is only observable with a hand-built double whose key object returns a different value from a second `.lower()` call. Happy to collapse the second normalisation if you would rather not carry it.
- Phases 2–4 — the cookies, meta-tag and asset layers, confidence scoring, and the proposed output schema — are untouched here, as are the HTTP fallback, body-size limit and redirect-loop handling listed under Additional Improvements.

## Verification

Beyond the suite: an independent reviewer ran a differential against the pre-refactor implementation over 20 paired responses served from a local raw-socket HTTP server — redirect chains, chunked transfer, `stream=True`, gzip/deflate/brotli, charset mismatches and undeclared charsets, invalid UTF-8, repeated and case-varied headers, empty and 2 MiB bodies, non-2xx, and real HEAD/204/304 — comparing the full observation including technology key and list order. All 20 compared equal. The signature tables were separately compared entry-by-entry against the base by AST, confirming all 27 technologies and 53 markers identical and in the same order.

Generated by Claude Opus 5 (implementation, testing, review), Kimi K3 (implementation, review), GPT-5.6 Luna (implementation), GPT-5.6 Sol (review, verification)

### DIFF
--- a/tests/test_fingerprint_engine.py
+++ b/tests/test_fingerprint_engine.py
@@ -1,0 +1,77 @@
+import unittest
+
+from tools.fingerprint import fingerprint
+from tools.fingerprint.engine import headers_layer, html_layer
+from tools.fingerprint.signatures import (
+    ANALYTICS_SIGNATURES,
+    CDN_SIGNATURES,
+    CMS_SIGNATURES,
+    JS_SIGNATURES,
+)
+
+
+class TestFingerprintEngine(unittest.TestCase):
+    """Contract of the I/O-free fingerprinting seam.
+
+    fingerprint() owns the normalisation (lowercased header names, lowercased
+    body) that the tool used to do inline, so callers hand it raw response
+    headers and raw response text.
+    """
+
+    def test_fingerprint_returns_the_full_technology_map(self):
+        result = fingerprint(
+            {"Server": "nginx/1.18", "X-Powered-By": "PHP/8.1", "CF-RAY": "abc123"},
+            '<LINK HREF="/WP-CONTENT/X.CSS"><SCRIPT SRC="/_NEXT/STATIC/A.JS"></SCRIPT>'
+            '<SCRIPT SRC="HTTPS://GOOGLE-ANALYTICS.COM/GA.JS"></SCRIPT>',
+        )
+
+        self.assertEqual(
+            result,
+            {
+                "web_server": "nginx/1.18",
+                "powered_by": "PHP/8.1",
+                "cdn": ["Cloudflare"],
+                "cms": ["WordPress"],
+                "javascript_frameworks": ["Next.js"],
+                "analytics": ["Google Analytics"],
+            },
+        )
+        self.assertEqual(
+            list(result.keys()),
+            [
+                "web_server",
+                "powered_by",
+                "cdn",
+                "cms",
+                "javascript_frameworks",
+                "analytics",
+            ],
+        )
+
+    def test_fingerprint_omits_categories_that_match_nothing(self):
+        self.assertEqual(fingerprint({}, ""), {})
+        self.assertEqual(fingerprint({"Server": "nginx"}, ""), {"web_server": "nginx"})
+
+    def test_layers_are_independent(self):
+        self.assertEqual(
+            headers_layer({"Server": "nginx", "CF-Ray": "x"}),
+            {"web_server": "nginx", "cdn": ["Cloudflare"]},
+        )
+        self.assertEqual(html_layer("WP-CONTENT"), {"cms": ["WordPress"]})
+        self.assertEqual(headers_layer({}), {})
+        self.assertEqual(html_layer(""), {})
+
+    def test_signature_tables_are_intact(self):
+        self.assertEqual(len(CDN_SIGNATURES), 6)
+        self.assertEqual(len(CMS_SIGNATURES), 7)
+        self.assertEqual(len(JS_SIGNATURES), 8)
+        self.assertEqual(len(ANALYTICS_SIGNATURES), 6)
+        self.assertEqual(CDN_SIGNATURES["Cloudflare"], ["cf-ray", "cf-cache-status"])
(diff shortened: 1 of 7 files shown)


tests/test_fingerprint_engine.py
```diff
@@ -0,0 +1,77 @@
...
+    def test_signature_tables_are_intact(self):
+        self.assertEqual(len(CDN_SIGNATURES), 6)
+        self.assertEqual(len(CMS_SIGNATURES), 7)
+        self.assertEqual(len(JS_SIGNATURES), 8)
+        self.assertEqual(len(ANALYTICS_SIGNATURES), 6)
+        self.assertEqual(CDN_SIGNATURES["Cloudflare"], ["cf-ray", "cf-cache-status"])
+        self.assertEqual(
+            ANALYTICS_SIGNATURES["Google Analytics"],
+            ["google-analytics.com", "gtag(", "ga("],
+        )
```
gaoharimran29-glitch, line 73: **Test naming**: `test_signature_tables_are_intact` currently checks the table sizes plus two representative entries rather than the full table contents. Since the characterization tests independently cover all 53 markers, the overall coverage is fine, but could we rename this to something like `test_signature_tables_have_expected_shape` to make the scope of this test more precise?

tests/test_tech_stack_detect.py
```diff
@@ -69,10 +97,395 @@ def test_no_security_analysis_in_output(self, mock_get):
...
+        # positives. They are pinned here so that fixing them is a visible,
+        # separately reviewed behaviour change rather than a silent side
+        # effect of moving the tables into the fingerprinting engine.
+        saga = self._detect(html="the saga(s) of vega(1)", headers={})
+        self.assertEqual(saga["technologies"], {"analytics": ["Google Analytics"]})
+
+        # "wix.com" matches inside an unrelated word, and Segment's
+        # "analytics.js" matches any script with that filename.
+        prose = self._detect(html="phoenix.community and /js/analytics.js", headers={})
+        self.assertEqual(prose["technologies"], {"analytics": ["Segment"]})
```
gaoharimran29-glitch, line 437: **Minor test/comment mismatch**: The comment mentions the wix.com substring false positive, but this particular fixture (phoenix.community and /js/analytics.js) doesn't exercise a Wix signature. Is the Wix case intentionally documented here without being pinned by this test, or should the fixture include a wix.com substring? If the latter, a small dedicated assertion would make the characterization explicit.

gaoharimran29-glitch (approved):
## Review Summary

Reviewed the refactor across the original `tech_stack_detect` implementation, the new `tools/fingerprint/` engine, signature tables, and the associated characterization and seam tests.

Overall, this looks good to merge. The extraction maintains a clean separation of responsibilities: `tech_stack_detect` remains responsible for HTTP I/O, response-state handling, error handling, and response assembly, while the fingerprinting engine encapsulates the technology matching logic behind an I/O-free interface. Separating the signature tables from the matching implementation also provides a clean extension point for the subsequent fingerprinting layers.

The characterization coverage is particularly strong for a behavior-preserving refactor. It covers the individual signature markers, technology/category ordering, case normalization, preservation of header values, known false-positive behavior, error paths, response-state/read ordering, and delegation through the new fingerprinting seam. The engine-level tests additionally validate layer independence and the expected signature-table structure.

I've left two minor inline comments around the test suite for clarification/precision. Neither is a blocking implementation issue. Please take a look at them when convenient. If either of these points warrants a test adjustment or cleanup, feel free to address it in a follow-up PR rather than holding up this behavior-preserving extraction.

No blocking issues found from my review.

**Approved.**

PR #172, opened 2026-08-20
fix(dns): resolve over UDP first and let TCP be the fallback (#144 item 3)

Nitjsefnie:
## Closes

Nothing on its own — this is **item 3 only** of the enhancement list, so the issue stays open for the remaining items. Refs #144

## Summary

Stops forcing TCP on DNS resolution. The three resolution paths in `dns_enumeration` now let dnspython use its normal UDP-first behaviour, with TCP still available as the fallback it already handles on truncation. Resolver lifetimes and the tool's public result shape are unchanged.

## What changed

- `tools/dns_tool.py` — dropped `tcp=True` from the three `resolver.resolve(...)` call sites: the ordinary record loop, the SRV query path, and the A/AAAA/CNAME subdomain probes. Six lines, nothing else.
- `tests/test_dns_enumeration.py` — the existing assertions **pinned the old behaviour** (`assert_any_call(..., tcp=True)`), so they had to be retargeted rather than added to. They still assert the lifetimes (`5` for records and SRV, `3` for subdomain probes) and the same mocked answers and exceptions.
- Added `test_udp_first_preserves_result_shape_and_content`, which asserts the exact call contract for all three families and pins the full returned structure — keys, per-type error strings, TTL map, resolver metadata and `subdomains_found` — so this refactor is provably behaviour-preserving apart from transport selection.

## Notes

- **TCP fallback was verified from source rather than assumed**, since it is the whole justification for the change. Against the pinned `dnspython==2.8.0`: `resolve()` defaults `tcp=False` (`dns/resolver.py:1251`), the UDP receive path raises `dns.message.Truncated` on the TC bit (`dns/query.py:872-877`), and `_Resolution.query_result` sets `retry_with_tcp` which `next_nameserver` turns into a TCP attempt against the same nameserver (`dns/resolver.py:808-813`, `761-766`).
- **Known limitation of the tests, disclosed rather than hidden.** The assertions pin exact calls, so reintroducing `tcp=True` at any one of the three sites fails the suite — verified by doing it three times. But a *partial* reintroduction hides: forcing TCP for only a non-asserted record type (say `MX`), or only a non-asserted subdomain, leaves the suite green. Call multiplicity is also unpinned, so a change that doubled every lookup would not be caught. A single sweep over `call_args_list` asserting that no call anywhere carries `tcp=` would close all three; I left it out to keep this diff to item 3, and it is probably better as its own change.
- **Sibling occurrence, out of scope here:** `tools/subdomain_takeover_tool.py:69` still passes `tcp=True` on its CNAME lookup. #154 covers that tool's error handling but not its transport. Flagging it rather than folding it in, since it belongs to that tool's issue and not to this one.
- Untouched by design: exception handling (item 1), the subdomain wordlist (item 5), timeouts, record types, and every key of the returned dictionary.

## Verification

- [ ] Updated Documentation ( if required ) — not required; no user-visible behaviour or output shape changed

Beyond the suite: an independent reviewer reintroduced `tcp=True` at each of the three call sites in turn, confirmed each produced an `AssertionError` (never an import or collection error) and that the suite returned green after each restore; confirmed the mocked resolver is genuinely exercised (208 calls in the main test, matching 8 record types + 5 SRV services + 65 subdomains × 3 types) so no assertion passes vacuously; and drove the base and changed implementations over identical mocked inputs across six scenarios — full success with TTLs, NXDOMAIN, an error matrix, a TXT decode error, the CAA raw fallback, and invalid input — comparing serialized output for exact equality. Zero mismatches.

Generated by GPT-5.6 Sol (analysis), GPT-5.6 Luna (implementation, testing), Kimi K3 (review, verification), Claude Opus 5 (brief, review)

### DIFF
--- a/tests/test_dns_enumeration.py
+++ b/tests/test_dns_enumeration.py
@@ -102,8 +102,88 @@ class TestDnsEnumeration(unittest.TestCase):
         self.assertEqual(result["errors"]["CAA"], "NoAnswer")
         self.assertIn("subdomains_found", result)
         self.assertEqual(resolver.nameservers, ["1.1.1.1", "8.8.8.8"])
-        resolver.resolve.assert_any_call("example.com", "TXT", lifetime=5, tcp=True)
-        resolver.resolve.assert_any_call("www.example.com", "A", lifetime=3, tcp=True)
+        resolver.resolve.assert_any_call("example.com", "TXT", lifetime=5)
+        resolver.resolve.assert_any_call("www.example.com", "A", lifetime=3)
+
+    @patch("tools.dns_tool.dns.resolver.Resolver")
+    def test_udp_first_preserves_result_shape_and_content(self, mock_resolver_class):
+        import dns.resolver as real_dns
+
+        resolver = Mock()
+        mock_resolver_class.return_value = resolver
+
+        def side_effect(name, rtype, lifetime=5, tcp=False):
+            if name == "example.com" and rtype == "A":
+                return _ResolverAnswer(["192.0.2.10"], 60)
+            if name == "_sip._tcp.example.com" and rtype == "SRV":
+                record = Mock()
+                record.priority = 10
+                record.weight = 20
+                record.port = 5060
+                record.target = "sip.example.com."
+                return [record]
+            if name == "www.example.com" and rtype == "A":
+                return _ResolverAnswer(["192.0.2.20"], 30)
+            raise real_dns.NoAnswer
+
+        resolver.resolve.side_effect = side_effect
+
+        result = dns_enumeration("example.com")
+
+        self.assertEqual(
+            result,
+            {
+                "success": True,
+                "domain": "example.com",
+                "errors": {
+                    "AAAA": "NoAnswer",
+                    "MX": "NoAnswer",
+                    "NS": "NoAnswer",
+                    "TXT": "NoAnswer",
+                    "CNAME": "NoAnswer",
+                    "SOA": "NoAnswer",
+                    "CAA": "NoAnswer",
+                },
+                "records": {
+                    "A": ["192.0.2.10"],
+                    "AAAA": [],
+                    "MX": [],
+                    "NS": [],
+                    "TXT": [],
+                    "CNAME": [],
+                    "SOA": [],
+                    "CAA": [],
+                },
+                "srv_records": {
+                    "_sip._tcp": [
+                        {
+                            "priority": 10,
+                            "weight": 20,
+                            "port": 5060,
+                            "target": "sip.example.com",
+                        }
+                    ],
+                    "_ldap._tcp": [],
+                    "_xmpp-client._tcp": [],
+                    "_kerberos._tcp": [],
+                    "_autodiscover._tcp": [],
+                },
+                "srv_errors": {},
+                "subdomains_found": ["www.example.com"],
+                "subdomain_errors": {},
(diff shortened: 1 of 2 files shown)


tests/test_dns_enumeration.py
```diff
@@ -299,9 +379,9 @@ def side_effect(domain, rtype, lifetime=5, tcp=False):
         self.assertNotIn("ftp.example.com", result["subdomains_found"])
         self.assertNotIn("www.example.com", result["subdomain_errors"])
         self.assertEqual(result["subdomains_found"].count("www.example.com"), 1)
-        resolver.resolve.assert_any_call("www.example.com", "A", lifetime=3, tcp=True)
-        resolver.resolve.assert_any_call("www.example.com", "AAAA", lifetime=3, tcp=True)
-        resolver.resolve.assert_any_call("mail.example.com", "CNAME", lifetime=3, tcp=True)
+        resolver.resolve.assert_any_call("www.example.com", "A", lifetime=3)
+        resolver.resolve.assert_any_call("www.example.com", "AAAA", lifetime=3)
+        resolver.resolve.assert_any_call("mail.example.com", "CNAME", lifetime=3)
```
gaoharimran29-glitch: **Blocking** :-  Assert the transport contract for all resolver calls

These assertions correctly verify the expected call arguments for the three representative subdomain probes, but they only prove that these specific calls occurred. They do not verify the transport argument for the other resolver calls made during the same `dns_enumeration()` execution.

Since this PR changes the resolver contract from explicitly forcing TCP to using dnspython's UDP-first behavior, could we make this test assert the transport setting across `resolver.resolve.call_args_list` as well? The regression should verify that **every resolver invocation made by `dns_enumeration()` uses `tcp=False` / does not explicitly force `tcp=True`**, rather than checking only these three representative calls.

This would cover the remaining A/AAAA/CNAME subdomain probes, all record-type lookups, and all SRV lookups without adding individual `assert_any_call()` assertions for each one.

Also, since the production code now omits `tcp` entirely and relies on dnspython's default, the existing tests that still explicitly assert `tcp=False` should be updated to remove that argument. The dedicated regression test can then own the transport invariant, while the other tests remain focused on their respective functional behavior.

The assertion should be against the application's calls to `resolver.resolve()`; dnspython's internal UDP → TCP retry on a truncated response is separate and should not be treated as a violation of this invariant.

gaoharimran29-glitch (approved):
## Review Summary

**Approve**

I reviewed the complete `tools/dns_tool.py` and `tests/test_dns_enumeration.py` within the scope of this PR.

The production-side change looks correct and appropriately scoped. The three `resolver.resolve()` paths now rely on dnspython's default UDP-first behavior instead of explicitly forcing TCP, while the existing lifetimes, DNS record handling, SRV/subdomain resolution, exception handling, TTL extraction, and result structure remain unchanged.

The new regression coverage also validates the representative resolution paths, and the remaining coverage limitation around non-asserted resolver calls is explicitly documented in the PR description along with the proposed `call_args_list` hardening. Given that this is intentionally scoped to item 3, I don't think that limitation needs to block this PR.

### Separate occurrence

Regarding the remaining `tcp=True` occurrence in `tools/subdomain_takeover_tool.py`, I'll verify that independently and update/create the relevant issue with the appropriate context. Once it is tracked, you or anyone else can pick it up and open a separate PR for it rather than expanding this change's scope.

Thanks for your contribution!

PR #173, opened 2026-08-20
fix(subdomain-takeover): report an incomplete probe as unknown, not safe (#154 item 10)

Nitjsefnie:
## Closes

Nothing on its own — this is **item 10**, plus only the probe-failure half of **item 3**, so the issue stays open for the DNS and unsupported-service halves and everything else on the list. Refs #154

## Summary

A subdomain-takeover probe that never completed was being reported as `safe`. In a security tool that is the dangerous direction of wrong: "could not check" and "checked, nothing found" collapsed to the same answer. Probe failures now land in a separate `unknown` bucket, carrying the per-scheme errors that explain why.

## What changed

- `tools/subdomain_takeover_tool.py`
  - `_probe()` still tries HTTPS then HTTP, but now returns the failure evidence instead of discarding it — per-scheme exception type and message.
  - `_confirms_takeover()` returns a tri-state (`confirmed` / `no_indicator` / `unable_to_probe`) instead of a bare bool, so the caller can tell the two false cases apart.
  - The caller routes only `unable_to_probe` to the new bucket. `confirmed` → `vulnerable` and `no_indicator` → `safe` behave exactly as before.
- `tests/test_subdomain_takeover.py` — the existing test **asserted the defect** (`safe == ["app.example.com"]` while both schemes raised `ConnectionError`), so it was retargeted rather than added to. New coverage: both schemes failing, a completed probe with no indicator, a completed probe with an indicator, and a mixed run asserting the three buckets are disjoint and total.

## Notes

- **This changes the return contract, and I would rather say so plainly than have you find it.** `unknown` is a new top-level key, and subdomains that used to inflate `safe` no longer appear there. Concretely: `len(vulnerable) + len(safe) == subdomains_checked` **no longer holds**, and any consumer treating `len(safe)` as "all clear" would now be reading a smaller number for the right reason. `subdomains_checked` still counts everything. If you would rather keep the old invariant and expose the state differently — a per-entry status instead of a third list, say — I am happy to reshape it.
- **The same "could not check → reported safe" pattern still exists one stage earlier, and this PR does not fix it.** `_resolve_cname()` swallows `Exception`, so a DNS timeout or SERVFAIL on an individual subdomain collapses to "no CNAME" and lands in `safe`. That is the DNS half of item 3 and it needs the specific-exception work from item 1, so I have deliberately left it alone rather than half-doing two items in one diff.
- Known test-coverage gap, disclosed: no test asserts `subdomains_checked` in a run that produces `unknown` entries, so a regression that undercounted it would stay green. The production value is correct (`len(subdomains)`); it is one assertion short of being pinned.
- `unknown` records carry `subdomain`, `reason` and `probe_errors`. They deliberately omit the matched `cname`/`service` — say the word if you would find those useful for triage and I will add them.

## Verification

- [ ] Updated Documentation ( if required ) — flagging instead: the return-shape change above is the documentation-relevant part, and I would rather you decide where it belongs than guess

Beyond the suite, an independent reviewer checked the direction that would actually matter if this went wrong. It drove the pre-change and post-change implementations side by side over a 42-scenario matrix — all seven fingerprints × three body casings × HTTPS-succeeds and HTTPS-fails-then-HTTP-succeeds — comparing full `vulnerable` records and `total_vulnerable`: **zero mismatches**, so no input that was flagged vulnerable before is anything other than vulnerable now. It verified the three buckets partition the input set exactly, using its own driver rather than the PR's tests; confirmed the invalid-domain and enumeration-failure responses are byte-identical to before; reintroduced each of the four collapse points singly and confirmed each turns the suite red with an `AssertionError`; and re-ran the whole test file with `socket.socket` and `socket.create_connection` replaced by raisers to prove no test touches the network.

Generated by GPT-5.6 Sol (analysis), GPT-5.6 Luna (implementation, testing), Kimi K3 (review, verification), Claude Opus 5 (brief, review)

### DIFF
--- a/tests/test_subdomain_takeover.py
+++ b/tests/test_subdomain_takeover.py
@@ -64,6 +64,7 @@ def test_vulnerable_subdomain_is_flagged(mock_enum, mock_resolver_class, mock_ge
     assert result["vulnerable"][0]["severity"] == "HIGH"
     assert "reason" in result["vulnerable"][0]
     assert result["safe"] == ["www.example.com"]
+    assert result.get("unknown") == []
     mock_get.assert_called_once()
 
 
@@ -89,6 +90,7 @@ def test_fingerprint_match_without_indicator_is_safe(mock_enum, mock_resolver_cl
     assert result["vulnerable"] == []
     assert result["total_vulnerable"] == 0
     assert result["safe"] == ["blog.example.com"]
+    assert result.get("unknown") == []
 
 
 @patch("tools.subdomain_takeover_tool.requests.get")
@@ -164,6 +166,7 @@ def test_aggregate_counts(mock_enum, mock_resolver_class, mock_get):
     assert result["vulnerable"][0]["subdomain"] == "dev.example.com"
     assert result["vulnerable"][0]["service"] == "GitHub Pages"
     assert result["safe"] == ["blog.example.com", "www.example.com"]
+    assert result.get("unknown") == []
 
 
 @patch("tools.subdomain_takeover_tool.requests.get")
@@ -189,6 +192,7 @@ def test_azure_vulnerable_matches_body(mock_enum, mock_resolver_class, mock_get)
     assert result["total_vulnerable"] == 1
     assert result["vulnerable"][0]["subdomain"] == "app.example.com"
     assert result["vulnerable"][0]["service"] == "Azure"
+    assert result.get("unknown") == []
     # HTTPS is attempted first and succeeded, so exactly one request is made to https://.
     assert mock_get.call_count == 1
     first_url = urlparse(mock_get.call_args_list[0].args[0])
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
APPROVE

This is a sound, appropriately scoped fix for item 10. Failed HTTPS/HTTP probes now remain distinguishable from completed negative probes and are reported as `unknown` with useful per-scheme error context, rather than incorrectly inflating `safe`.

The tests cover the intended tri-state behavior and verify that the result buckets are disjoint and exhaustive for probed subdomains.

The return-shape change is clearly called out and preserves existing semantics for completed probes.

Please open a separate, fully tested follow-up PR for the remaining item 3 paths: DNS-resolution errors and unsupported services must also resolve to `unknown`, not `safe`.

Issue #143, opened 2026-08-03
Docs: Add dedicated documentation for each reconn tool


gaoharimran29-glitch:
## Summary

AynOps currently provides 16 MCP tools covering reconnaissance tools. While the README gives a high-level overview, there is no dedicated documentation explaining each tool in detail.

Adding individual documentation pages for every tool would make the project easier to use, improve discoverability, and provide contributors with a single source of truth for tool behavior.

## Proposed Structure

```text
docs/
└── tools/
    ├── README.md
    ├── whois_lookup.md
    ├── dns_enumeration.md
    ├── port_scan.md
    ├── ssl_inspect.md
    ├── email_security_check.md
    ├── tech_stack_detect.md
    ├── cert_transparency.md
    ├── asn_lookup.md
    ├── ip_reputation.md
    ├── full_recon.md
    ├── headers_analyzer.md
    ├── cve_lookup.md
    ├── cloud_exposure_check.md
    ├── trace_redirects.md
    ├── robots_txt_inspect.md
    └── hibp_check.md
```

## Each documentation page should include

- Overview
- Purpose and common use cases
- Parameters
- Optional parameters ( if applicable; explicitly state when there are none) 
- Example MCP request
- Example successful response
- Response field descriptions
- Errors and timeouts
- Notes and limitations
- API key requirements ( if applicable; explicitly state when there are none) 
- Related tools

## Documentation Index

- Create a `docs/tools/README.md` that serves as an index for all available tools.
- Add a link to the documentation index from the main `README.md`.
- Update the tool tables in the main `README.md` so that **each tool name is a clickable link** pointing to its corresponding documentation page (e.g., `whois_lookup` → `docs/tools/whois_lookup.md`).
- Ensure any future tools added to the project are also linked from both the documentation index and the main `README.md`.

### Included in `full_recon`

- [x] whois_lookup
- [x] dns_enumeration
- [x] port_scan
- [x] ssl_inspect
- [ ] email_security_check
- [ ] tech_stack_detect
- [ ] cert_transparency
- [ ] asn_lookup
- [ ] ip_reputation
- [ ] full_recon

### Standalone Tools

- [ ] headers_analyzer
- [ ] cve_lookup
- [ ] cloud_exposure_check
- [ ] trace_redirects
- [ ] robots_txt_inspect
- [ ] hibp_check

## Contribution Notes

[...]

PR #175, opened 2026-08-24
docs(whois): add WHOIS tool documentation

Nitjsefnie:
## Closes

Relates to #143

## Summary

Documents the current `whois_lookup` contract and introduces a complete tools index so users can find the new reference while the remaining issue scope stays open.

## What changed

- Added `docs/tools/whois_lookup.md` with input normalization and validation, an MCP request example, response fields and value shapes, timeout and error outcomes, limitations, and related tools.
- Added `docs/tools/README.md`, grouping and enumerating all 17 tools currently registered by the server.
- Linked the existing `whois_lookup` row in the root README to its dedicated page.

## Notes

- This is the first documentation slice related to #143; dedicated pages for the remaining tools are intentionally outside this PR.
- No runtime code, dependencies, or test behavior changed.

## Verification

  - Focused WHOIS regression tests: `python3 -m pytest tests/test_whois_lookup.py -q` (`7 passed`).
  - Relative-link checks resolved all four checked links, and the inventory check matched all 17 registered tools.
- [ ] Added/updated tests for the changed behavior (documentation-only change; no runtime behavior changed)
- [x] All tests pass (`pytest` via fork wrapper run `32688816381`)
  - The run had one successful `test` job with exact provenance `38042e7abe717aa8d8e9dfe5a46c861ab4441d85` and reported `366 passed, 85 subtests passed`.
  - Independent review approved exact SHA `38042e7abe717aa8d8e9dfe5a46c861ab4441d85` for shipping.

Generated by GPT-5.6 Luna (implementation, testing, verification), GPT-5.6 Sol (brief, review, verification)

### DIFF
--- a/docs/tools/whois_lookup.md
+++ b/docs/tools/whois_lookup.md
@@ -23,9 +23,9 @@ The MCP tool exposes one required input:
 There are no optional parameters. The server strips surrounding whitespace,
 converts the value to lowercase, and removes trailing dots before validation.
 Validation requires at least one dot, labels of at most 63 characters, a total
-length of at most 253 characters, and a final label containing at least two
-ASCII letters. Each non-final label may contain letters, digits, and internal
-hyphens, but may not start or end with a hyphen.
+length of at most 253 characters, and a final label consisting entirely of at
+least two ASCII letters. Each non-final label may contain letters, digits, and
+internal hyphens, but may not start or end with a hyphen.
 
 ## Example MCP request
 
@@ -79,7 +79,7 @@ On success, the response contains `success: true` and the following fields:
 | `success` | boolean | `true` for a completed WHOIS lookup. |
 | `domain` | string, list of strings, or null | The `domain_name` value returned by the WHOIS library. A single parsed value remains a string; multiple values are returned as a list. |
 | `registrar` | string or null | Registrar returned by the WHOIS response. |
-| `registrar_url` | string or null | Registrar URL, when supplied. |
+| `registrar_url` | string, list of strings, or null | Registrar URL, when supplied. A single parsed value is a string; repeated upstream values are returned as a list. |
 | `whois_server` | string or null | WHOIS server returned by the response. |
 | `creation_date` | string, list of strings, or null | Creation date. Date objects are converted to strings; multiple values are returned as a list. |
 | `expiration_date` | string, list of strings, or null | Expiration date, with the same conversion and list behavior as `creation_date`. |
--- a/docs/tools/whois_lookup.md
+++ b/docs/tools/whois_lookup.md
@@ -6,13 +6,13 @@
 returns the fields that the upstream WHOIS response provides. It does not
 require an API key.
 
-## Common uses
+## Purpose and common use cases
 
 Use this tool to inspect registration metadata such as the registrar, registry
 dates, name servers, registration status, and the organization associated with
 a domain. WHOIS responses vary by registry, so some fields may be unavailable.
 
-## Input
+## Parameters
 
 The MCP tool exposes one required input:
 
@@ -20,12 +20,14 @@ The MCP tool exposes one required input:
 |---|---|---|---|
 | `domain` | string | Yes | A fully qualified domain name to query. IP addresses, `localhost`, bare labels, and malformed domain names are rejected. |
 
-There are no optional parameters. The server strips surrounding whitespace,
-converts the value to lowercase, and removes trailing dots before validation.
 Validation requires at least one dot, labels of at most 63 characters, a total
 length of at most 253 characters, and a final label consisting entirely of at
 least two ASCII letters. Each non-final label may contain letters, digits, and
 internal hyphens, but may not start or end with a hyphen.
+
+## Optional parameters
+
+None.
 
 ## Example MCP request
 
@@ -37,7 +39,7 @@ internal hyphens, but may not start or end with a hyphen.
   "params": {
     "name": "whois_lookup",
     "arguments": {
-      "domain": " Example.COM. "
+      "domain": "example.com"
     }
   }
 }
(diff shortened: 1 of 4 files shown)


docs/tools/whois_lookup.md
```diff
@@ -0,0 +1,127 @@
...
+
+```json
+{
+  "jsonrpc": "2.0",
+  "id": 1,
+  "method": "tools/call",
+  "params": {
+    "name": "whois_lookup",
+    "arguments": {
+      "domain": " Example.COM. "
```
gaoharimran29-glitch: The current value is `"domain": " Example.COM. "` with leading/trailing whitespace. While this works because `normalize_domain()` strips surrounding whitespace, I don't think this is ideal for the **standard MCP request example**. The example should preferably use a canonical input such as:

`"domain": "example.com"`

If we want to demonstrate that normalization accepts whitespace, casing differences, or a trailing dot, that behavior can be mentioned separately in the notes/limitations rather than making the primary request example look unusual.

docs/tools/whois_lookup.md
```diff
@@ -0,0 +1,127 @@
...
+
+## Input
+
+The MCP tool exposes one required input:
+
+| Name | Type | Required | Description |
+|---|---|---|---|
+| `domain` | string | Yes | A fully qualified domain name to query. IP addresses, `localhost`, bare labels, and malformed domain names are rejected. |
+
+There are no optional parameters. The server strips surrounding whitespace,
```
gaoharimran29-glitch: Please explicitly mention that there are no optional parameters.

Since the tool currently only accepts the required domain parameter, the `Optional parameters section` can simply state that there are none. This keeps the page consistent with the standard template.

gaoharimran29-glitch (changes requested):
Thanks for the detailed documentation. I reviewed this against the current `whois_lookup` implementation, including the normalization/validation helpers, response construction, timeout handling, and exception paths. Overall, the documented behavior is well aligned with the implementation.

I’ve left inline comments for the non-canonical whitespace in the MCP request example and for explicitly documenting that there are no optional parameters.

One additional consistency point: please explicitly include the `API key requirements` section and state `None required` for this tool. I’ve updated the issue requirements to make this explicit for every tool. The original wording said `where applicable`, which was ambiguous on my side and could reasonably be interpreted as allowing the section to be omitted when no API key is required. I’ve clarified this so every tool page follows the same schema.

I’ve also updated the issue description to make the documentation contract more explicit, including the standardized sections and the `Errors and timeouts` section. Please align this page with those exact section names rather than introducing tool-specific variants such as `Input` or `Common uses`.

Even before these issue-description updates, the original issue explicitly asked contributors to follow a **consistent documentation format**. Some of the headings in this contribution already diverge from the terminology specified there, so I’d still recommend aligning them with the standardized naming rather than treating the updated requirements as a new constraint introduced after this contribution.

The goal here is to establish a stable documentation schema across all tools so that future contributions remain structurally consistent, searchable, and easy to compare/review. The content of each section can vary based on the tool, but the section naming and ordering should remain consistent.

Once the inline points are addressed and the page is aligned with the updated issue structure, I think this will be in good shape.

docs/tools/whois_lookup.md
Nitjsefnie: Updated the standard request to use the canonical `example.com` value, with accepted whitespace, casing, and trailing-dot normalization documented under Notes and limitations. The change is pushed and verified at `86cd592b049048554a217cd47cc9e213df8b9a86`.

docs/tools/whois_lookup.md
Nitjsefnie: Added a dedicated Optional parameters section and stated explicitly that there are none. The change is pushed and verified at `86cd592b049048554a217cd47cc9e213df8b9a86`.

tests/test_whois_lookup.py
```diff
@@ -78,5 +79,27 @@ def test_whois_none_dates_return_null(self, mock_whois):
...
+            repo_root / "docs/tools/README.md",
+        )
+        for summary_path in summary_paths:
+            summary = next(
+                line
+                for line in summary_path.read_text(encoding="utf-8").splitlines()
+                if "whois_lookup" in line and "whois_lookup.md" in line
+            )
+            self.assertIn("registrant organization", summary.lower())
+            self.assertNotIn("owner", summary.lower())
```
gaoharimran29-glitch: Could you clarify the rationale for adding this test? The issue focuses on documenting and standardizing the tool documentation, but doesn't require automated assertions against README/documentation wording. The `org` vs `owner` assertions against specific README lines also couple the test to documentation structure rather than testing tool behavior.

Unless there is an existing bug or explicit requirement to enforce this terminology, I think the documentation-content assertions are outside the scope of this PR and could be removed.

gaoharimran29-glitch (changes requested):
Thanks for addressing the previous review feedback and aligning the documentation with the updated requirements. The implementation and documentation changes look good overall.

I have one remaining concern regarding the newly added documentation-content test. Please see the inline comment for the rationale/scope of that test.

Other than that, the documentation structure and README integration now look aligned with the updated issue requirements.

tests/test_whois_lookup.py
Nitjsefnie: Removed the documentation-content test and its now-unused `Path` import while preserving the documentation correction. The updated head passed exact-SHA fork verification and both PR checks.

gaoharimran29-glitch (approved):
Reviewed the latest changes. The requested documentation updates have been addressed, including alignment with the standardized section structure, explicit handling of optional parameters and API-key requirements, correction of the MCP request example, and removal of the out-of-scope documentation-content test.

The tool documentation is now consistent with the updated issue requirements and the implementation details look aligned with the current `whois_lookup` behavior.

No further blocking concerns from my side. This is ready to merge.

PR #176, opened 2026-08-24
feat(dns): run independent lookups concurrently

Nitjsefnie:
## Closes

Relates to #144

## Summary

Runs independent DNS work through a bounded, category-fair executor while preserving the conclusive NXDOMAIN short-circuit and deterministic output semantics. This addresses item 2 only; the other issue items remain open.

## What changed

- Kept the first configured target-record lookup ahead of executor creation so a conclusive NXDOMAIN returns before unrelated work starts.
- Interleaved the remaining record, SRV, and common-subdomain tasks through a bounded executor so each category can make progress under production queue depths.
- Aggregated completed work in configured order, preserving prior TTL handling, error classification, parsing, and output ordering.
- Added synchronization regressions for production-depth category starvation and NXDOMAIN shutdown behavior.

## Notes

- The public function signature and result schemas are unchanged.
- This PR is limited to issue #144 item 2; the remaining items are intentionally outside its scope.

## Verification

  - The implementer report records assertion-first regressions for production-depth category starvation and NXDOMAIN shutdown.
  - The implementer report records both single-defect mutations being killed by their intended assertions.
- [x] All tests pass (`python3 -m unittest tests.test_dns_enumeration`: 22 tests)
  - Fork wrapper run `32691463503` had one successful `test` job with exact provenance `1cad4402eee2ac15de1dae0eea2cdc4aadc518d5` and reported `368 passed, 85 subtests passed`.
  - Independent task and final whole-branch reviews both returned `SHIP 1cad4402eee2ac15de1dae0eea2cdc4aadc518d5`.
- [ ] Updated Documentation (not required; public interfaces and result schemas are unchanged)

Generated by GPT-5.6 Sol (brief, implementation, testing, review, verification), GPT-5.6 Luna (verification)

### DIFF
--- a/tools/dns_tool.py
+++ b/tools/dns_tool.py
@@ -1,3 +1,5 @@
+from concurrent.futures import ThreadPoolExecutor
+
 import dns.exception
 import dns.name
 import dns.resolver
@@ -10,6 +12,7 @@ PUBLIC_RESOLVERS = ["1.1.1.1", "8.8.8.8"]
 RESOLVER_TIMEOUT = 2.0
 RESOLVER_LIFETIME = 5
 SUBDOMAIN_LIFETIME = 3
+MAX_CONCURRENT_LOOKUPS = 10
 
 # Record types enumerated for the target domain.
 RECORD_TYPES = ["A", "AAAA", "MX", "NS", "TXT", "CNAME", "SOA", "CAA"]
@@ -110,6 +113,19 @@ def _resolver_metadata(resolver) -> dict:
     }
 
 
+def _lookup_subdomain(resolver, full: str) -> tuple[bool, dict]:
+    errors = {}
+    for rtype in SUBDOMAIN_RECORD_TYPES:
+        try:
+            resolver.resolve(full, rtype, lifetime=SUBDOMAIN_LIFETIME)
+            return True, {}
+        except SUBDOMAIN_LOOKUP_ERRORS:
+            continue
+        except Exception as exc:
+            errors[rtype] = f"unexpected: {type(exc).__name__}"
+    return False, errors
+
+
 def dns_enumeration(domain: str) -> dict:
     """
     Enumerate DNS records for a domain.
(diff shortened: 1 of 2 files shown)


tools/dns_tool.py
```diff
@@ -129,109 +146,143 @@ def dns_enumeration(domain: str) -> dict:
...
+                        raise first_record_error
+                    answers = first_record_answers
                 else:
-                    records[rtype] = [str(r) for r in answers]
-            except UnicodeDecodeError as exc:
-                # TXT chunks and CAA values are arbitrary remote octets, not
-                # guaranteed UTF-8. Keep this record type's failure visible
-                # without swallowing defects in our own formatting logic.
+                    answers = record_futures[index - 1].result()
+            except dns.resolver.NXDOMAIN:
```
gaoharimran29-glitch, line 211: This is no longer an immediate short-circuit for later target-record lookups. By this point, the executor has already scheduled record/SRV/subdomain work, and returning from inside the `ThreadPoolExecutor` context invokes `shutdown(wait=True)`. Therefore, `AAAA → NXDOMAIN` can still block the caller on unrelated in-flight/queued DNS work, regressing the previous conclusive-failure semantics. Please decouple this failure path from synchronous executor shutdown by cancelling pending work and avoiding an unconditional wait on unrelated tasks.

gaoharimran29-glitch (changes requested):
### Request changes

The refactor introduces a later-record `NXDOMAIN` regression. While `A → NXDOMAIN` still short-circuits before the executor is created, `A → NoAnswer` followed by `AAAA → NXDOMAIN` occurs after unrelated record/SRV/subdomain work has already been scheduled.

Because the `return` executes inside the `ThreadPoolExecutor` context, `shutdown(wait=True)` can make the caller wait for unrelated DNS work before receiving the conclusive failure.

Please preserve the previous immediate-failure semantics for later target-record `NXDOMAIN`: cancel pending work where possible and avoid synchronously joining unrelated tasks on the failure path.

Please also add a regression test covering:

`A → NoAnswer → later target record → NXDOMAIN`, with unrelated work blocked, and assert that `dns_enumeration()` returns before that unrelated work is released.

tools/dns_tool.py
Nitjsefnie: Fixed in cd69127fb4a179659a4edc44be9fabb0facf7fa6: later target-record NXDOMAIN now cancels queued futures and shuts down without waiting for unrelated work. The new blocked-work regression proves the caller returns first, and this exact SHA passed 370 tests plus 85 subtests in fork CI.

gaoharimran29-glitch (approved):
## Review Summary: **Approved**

The latest revision resolves the previously identified **later-record NXDOMAIN lifecycle regression**.

* The failure path now explicitly transitions the executor into **cancellation-oriented shutdown** before returning.
* Pending futures are cancelled, while executor teardown no longer synchronously waits on unrelated DNS work.
* The caller receives the **conclusive target-domain failure** without being coupled to queued SRV/subdomain lookup completion.
* In-flight DNS requests remain non-cancellable, as expected for thread-backed work, but are correctly **detached from the request return path**.
* The new synchronization regression test covers the missing topology: initial A lookup → non-conclusive negative, later AAAA lookup → NXDOMAIN, unrelated worker occupied, and subsequent work queued.
* The test verifies that enumeration returns **before unrelated work is released** and that queued work is cancelled rather than started after the conclusive failure.

This restores the original **short-circuit semantics** while retaining bounded, category-fair concurrent execution and deterministic successful-path aggregation.

**No further blocking concerns from this review.**

PR #189, opened 2026-09-01
fix(headers): standardize disclosure header keys to hyphenated form

### LINKED ISSUE #187: Bug: Standardize header analysis result keys to use hyphenated HTTP header names
### Summary

The dict returned by `_analyze_raw_headers()` (and surfaced as `headers` / `redirect_chain[i]["analysis"]` in `headers_analyzer()`'s output) uses two different key-naming conventions for header-derived fields.

For the primary security headers, the header's native hyphenated form is used directly as the dict key:

```python
headers["strict-transport-security"] = {...}
headers["content-security-policy"] = {...}
headers["x-frame-options"] = {...}
headers["x-content-type-options"] = {...}
headers["referrer-policy"] = {...}
headers["permissions-policy"] = {...}
```

However, the information-disclosure loop at the bottom of the same function converts hyphens to underscores before inserting the key:

```python
for hdr in ("server", "x-powered-by", "x-aspnet-version", "x-generator"):
    val = raw_headers.get(hdr, "")
    if val:
        headers[hdr.lower().replace("-", "_")] = {...}
```

This produces keys such as `"x_powered_by"` and `"x_aspnet_version"` alongside `"x-frame-options"` and `"x-content-type-options"` in the same flat result dict.

### Why this matters

Consumers that programmatically iterate over the returned `headers` dict — such as dashboard rendering, scan diffing, or lookup-table generation — currently need to account for two different key formats instead of relying on one predictable convention.

This can also cause silent lookup failures. For example:

```python
headers.get("x-frame-options")
```

works as expected, while:

```python
headers.get("x-powered-by")
```

returns `None` even when the header is present, because the actual key is `"x_powered_by"`.

The inconsistency is also undocumented and appears unintentional, which makes it easy for downstream consumers or future contributors to handle these fields differently.

### Proposed fix

Standardize all header-derived keys on the hyphenated form, matching the actual HTTP header syntax and the existing convention used by the primary security headers.

The information-disclosure loop should therefore use:

```python
headers[hdr.lower()] = {...}
```

instead of:

```python
headers[hdr.lower().replace("-", "_")] = {...}
```

This keeps the output schema consistent without changing the underlying header values or analysis structure.

### Test coverage required

Update `tests/test_headers_tool.py` to cover the standardized key format.

Add tests that:

* Verify that a mocked response containing `server` and/or `x-powered-by` produces keys using the hyphenated convention, e.g. `"x-powered-by"`, rather than `"x_powered_by"`.
* Verify that a fully populated `_analyze_raw_headers()` result uses a consistent naming convention across all header-derived keys.
* Prefer a structural assertion over only hardcoding individual headers, so a future header-analysis addition cannot silently reintroduce mixed key formats.

For example, if the hyphenated convention is retained, the test should validate that header-derived keys do not contain underscore-based transformations of HTTP header names.

### Files to be touched

* `tools/headers_tool.py`
* `tests/test_headers_tool.py`

Ankur-Kataria:
## Closes

Closes #187

## Summary

`_analyze_raw_headers()` stored information-disclosure headers (`x-powered-by`, `x-aspnet-version`, `x-generator`) under underscore-converted keys while all security headers used their native hyphenated HTTP form. This removes the `.replace("-", "_")` so every key in the returned dict is consistently hyphenated.

## What changed

- `tools/headers_tool.py`: Remove `.replace("-", "_")` from the information-disclosure loop (one-char fix at line 485)
- `tests/test_headers_tool.py`: Add two regression tests
  - `test_disclosure_headers_use_hyphenated_keys` — verifies `x-powered-by` / `x-aspnet-version` appear under hyphenated keys and their underscore variants are absent
  - `test_disclosure_headers_all_use_hyphenated_keys` — covers all four disclosure headers together

## Notes

The `server` key was never affected (no hyphens to convert); only multi-word header names like `x-powered-by` produced the mixed-format output. No callers inside this repo were relying on the underscore form.

## Verification

- [x] All tests pass (`python -m pytest tests/test_headers_tool.py` → 53 passed)

### DIFF
--- a/tests/test_headers_tool.py
+++ b/tests/test_headers_tool.py
@@ -301,6 +301,39 @@ class TestHeadersAnalyzer(unittest.TestCase):
         self.assertIn("exposes technology", result["headers"]["server"]["issue"])
 
     @patch("tools.headers_tool.requests.get")
+    def test_disclosure_headers_use_hyphenated_keys(self, mock_get):
+        """x-powered-by and similar headers must appear under their native
+        hyphenated key, not an underscore-converted form (regression guard
+        for the hdr.lower().replace('-', '_') bug reported in issue #187)."""
+        mock_get.return_value = _resp(200, {
+            "X-Powered-By": "PHP/8.1",
+            "X-AspNet-Version": "4.0.30319",
+        })
+        result = headers_analyzer("example.com")
+        headers = result["headers"]
+        self.assertIn("x-powered-by", headers)
+        self.assertIn("x-aspnet-version", headers)
+        self.assertNotIn("x_powered_by", headers)
+        self.assertNotIn("x_aspnet_version", headers)
+        self.assertIn("exposes technology", headers["x-powered-by"]["issue"])
+
+    @patch("tools.headers_tool.requests.get")
+    def test_disclosure_headers_all_use_hyphenated_keys(self, mock_get):
+        """All four disclosure headers must appear under their native
+        hyphenated keys; no underscore-converted variants must be present."""
+        mock_get.return_value = _resp(200, {
+            "X-Powered-By": "Express",
+            "X-Generator": "Drupal 9",
+            "X-AspNet-Version": "4.0.30319",
+            "Server": "nginx",
+        })
+        result = headers_analyzer("example.com")
+        headers = result["headers"]
+        underscore_keys = {"x_powered_by", "x_aspnet_version", "x_generator"}
+        for bad_key in underscore_keys:
+            self.assertNotIn(bad_key, headers, f"Underscore-form key {bad_key!r} found")
+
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch:
Thank you again for taking the time to work on this and for the care you put into the fix and tests — the implementation is correct and appreciated.
I’m going to close this PR as superseded by #190, not because there is a problem with your solution. Both PRs make the required production change, but #190 was selected because it includes slightly broader direct coverage of `_analyze_raw_headers()` and updates the related internal documentation.
Please don’t take the closure as a negative reflection on the quality of your contribution. You moved quickly, kept the scope focused, and added useful regression coverage. I’d be very happy to see you pick up another open issue — contributions like this are genuinely welcome.

PR #190, opened 2026-09-01
fix(headers): standardize header analysis result keys to hyphenated format

### LINKED ISSUE #187: Bug: Standardize header analysis result keys to use hyphenated HTTP header names

HeaTTap:
## Closes

Closes #187

## Summary

Standardize header analysis result keys to use native hyphenated HTTP header names rather than converting to underscores in information disclosure headers.

## What changed

- Updated headers dict assignment in tools/headers_tool.py to use hdr.lower() instead of replacing hyphens with underscores
- Updated comment reference in tools/signals/headers.py
- Added unit tests in tests/test_headers_tool.py validating hyphenated keys for information disclosure headers and confirming no underscore keys exist

## Notes

None

## Verification

- [x] All tests pass (uv run pytest tests/ -v)
- [x] Checked CONTRIBUTING.md and applied all changes

### DIFF
--- a/tests/test_headers_tool.py
+++ b/tests/test_headers_tool.py
@@ -2,7 +2,7 @@ import unittest
 from unittest.mock import patch, Mock
 from curl_cffi.requests.errors import RequestsError
 from curl_cffi.requests.headers import Headers
-from tools.headers_tool import headers_analyzer
+from tools.headers_tool import _analyze_raw_headers, headers_analyzer
 
 
 def _resp(status_code: int, headers: dict, body: str = ""):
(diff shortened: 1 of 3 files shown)


gaoharimran29-glitch (approved):
Approved — this fully addresses #187.
The production change correctly removes the underscore normalization from the information-disclosure header path, so `_analyze_raw_headers()` now preserves the same lower-case, hyphenated naming convention used by the security-header analysis (x-powered-by, `x-aspnet-version`, and `x-generator`).
The test coverage is particularly solid:
- validates the public headers_analyzer() output;
- exercises all four disclosure headers together;
- directly tests `_analyze_raw_headers()` with a complete header set;
- adds a structural regression guard ensuring analyzed keys remain lowercase and contain no underscore-transformed header names.

Updating the related tools/signals/headers.py documentation also keeps the internal contract accurate. The scope is focused, CI is green, and I don’t see any changes needed before merge. Great work — approved.

PR #191, opened 2026-09-01
Fix CSP wildcard token matching

### LINKED ISSUE #186: Bug:  CSP wildcard detection uses substring matching instead of token matching, causing false positives/negatives
### Summary
`_analyze_raw_headers()` in `headers_tool.py` detects an unrestricted CSP wildcard source using a raw substring check:

```python
if "default-src *" in csp_lower or "script-src *" in csp_lower:
    issues.append("Wildcard (*) source defeats the purpose of CSP")
    severity = "high"
```

Because this is a plain substring test against the lowercased header value rather than a parse of CSP directive/source syntax, it produces both false positives and false negatives.

### False positive (confirmed)
A scoped wildcard source such as `script-src *.trusted-cdn.com` — a legitimate, reasonably restrictive pattern that limits script loading to a specific vendor's subdomains — contains the literal substring `"script-src *"` and gets misclassified as a full open wildcard at **high severity**.

Repro:
```python
>>> csp_lower = "script-src *.trusted-cdn.com"
>>> "script-src *" in csp_lower
True
```
Expected: no issue raised, or at most a lower-severity note about broad subdomain scoping. Actual: reported as "Wildcard (*) source defeats the purpose of CSP", severity `high`.

### False negative (also present)
Conversely, a directive with non-canonical but valid whitespace, e.g. `script-src  *` (double space) or `script-src * ;`, will not match the fixed literal `"script-src *"` and the wildcard will go undetected entirely.

### Root cause
CSP has a defined directive/token grammar (directives separated by `;`, each directive's value a whitespace-separated list of source expressions). Treating the header as an opaque string and pattern-matching against it cannot correctly distinguish "wildcard applies to the whole directive" from "wildcard is part of a scoped host expression," and is fragile to incidental whitespace variation.

### Proposed fix
Parse the CSP value properly before evaluating:
1. Split the header on `;` to get individual directives.
2. For each directive, split off the directive name and tokenize the remaining value on whitespace to get the list of source expressions.
3. Flag a directive only when `*` appears as a **standalone token** in that list (i.e. `token == "*"`), not when `*` appears as part of a larger token like `*.trusted-cdn.com`.
4. Apply this per relevant directive (`default-src`, `script-src`, and ideally extend to `style-src`, `img-src`, `connect-src` etc. rather than hardcoding just two — open to discussion on scope, can be a follow-up).

### Test coverage required
Add cases to the local test file covering at minimum:
- `script-src *.trusted-cdn.com` → should NOT raise the "wildcard defeats CSP" issue.
- `default-src *.example.com` → should NOT raise the issue.
- `script-src *` (standalone) → SHOULD raise the issue at high severity.
- `default-src *` (standalone) → SHOULD raise the issue at high severity.
- `script-src  *` (irregular whitespace) → SHOULD still raise the issue (regression guard for the false-negative side).
- A directive where `*` appears only in an unrelated directive (e.g. `img-src *` present but `script-src 'self'`) → confirm the issue is attributed only to the directive that actually contains it, not applied globally.

Files to be touched :- `tools\headers_tool.py` , `tests\test_headers_tool.py`

AnandkumarMall:
This PR resolves #186 by properly tokenizing the CSP directive value to check for a standalone `*` wildcard, avoiding false positives on scoped wildcards like `*.trusted.com` and false negatives on irregular whitespace.

### DIFF
--- a/tests/test_headers_tool.py
+++ b/tests/test_headers_tool.py
@@ -188,6 +188,34 @@ class TestHeadersAnalyzer(unittest.TestCase):
         self.assertEqual(csp["severity"], "high")
 
     @patch("tools.headers_tool.requests.get")
+    def test_csp_wildcard_source_scoped_not_flagged(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Content-Security-Policy": "script-src *.trusted-cdn.com; default-src *.example.com",
+        })
+        result = headers_analyzer("example.com")
+        csp = result["headers"]["content-security-policy"]
+        self.assertNotIn("Wildcard", csp["issue"])
+
+    @patch("tools.headers_tool.requests.get")
+    def test_csp_wildcard_irregular_whitespace_flagged(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Content-Security-Policy": "script-src   * ;",
+        })
+        result = headers_analyzer("example.com")
+        csp = result["headers"]["content-security-policy"]
+        self.assertIn("Wildcard", csp["issue"])
+        self.assertEqual(csp["severity"], "high")
+
+    @patch("tools.headers_tool.requests.get")
+    def test_csp_wildcard_in_unrelated_directive(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Content-Security-Policy": "img-src *; script-src 'self'",
+        })
+        result = headers_analyzer("example.com")
+        csp = result["headers"]["content-security-policy"]
+        self.assertNotIn("Wildcard", csp["issue"])
+
+    @patch("tools.headers_tool.requests.get")
     def test_csp_missing_default_src_flagged_once(self, mock_get):
         mock_get.return_value = _resp(200, {
             "Content-Security-Policy": "script-src 'self'",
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
Thanks for the contribution!

I reviewed the implementation and the accompanying regression coverage. The updated CSP wildcard detection correctly moves away from naive substring matching and evaluates directive source tokens instead. This addresses the false-positive case for scoped wildcards such as `*.trusted-cdn.com` while preserving detection of unrestricted `*` sources.

The added tests also provide good coverage across the relevant edge cases, including:

* Scoped domain wildcards not being reported as unrestricted wildcards.
* Irregular whitespace around wildcard sources.
* Wildcards appearing in unrelated directives.
* Preservation of the expected high-severity behavior for unrestricted wildcards.

The changes are scoped appropriately and directly address the reported issue without introducing unrelated modifications.

**Approved from my side. Thanks again for taking the time to contribute!** 🎉

If you're enjoying the project and would like to support its development, consider leaving the repository a ⭐ on GitHub. It genuinely helps with visibility and encourages further contributions!

PR #193, opened 2026-09-02
fix: respect fallback list ordering in Referrer-Policy header check

### LINKED ISSUE #192: Bug: Referrer-Policy check does not respect comma-separated fallback-list ordering, producing false negatives
### Summary

`_analyze_raw_headers`'s Referrer-Policy branch determines "safe vs. unsafe" via an unordered substring/membership scan over the raw header value. This is incorrect per spec: `Referrer-Policy` supports a comma-separated list of policy tokens as a legacy-fallback mechanism, and the **last token the client recognizes** is the one actually applied — not "any token present anywhere in the string." As written, the check can classify a header as safe (`severity: low`, `issue: None`) when the effective, browser-applied policy is actually unsafe.

### Current implementation

```python
rp = raw_headers.get("referrer-policy", "")
if rp:
    rp_lower = rp.lower().strip()
    good_policies = ["no-referrer", "strict-origin-when-cross-origin", "same-origin", "strict-origin"]
    if any(p in rp_lower for p in good_policies):
        headers["referrer-policy"] = {
            "present": True,
            "value": rp,
            "issue": "None",
            "severity": "low",
        }
    ...
```

### Root cause

`Referrer-Policy` allows multi-value headers of the form:

```
Referrer-Policy: <fallback-policy>, <preferred-policy>
```

This exists specifically so older browsers that don't recognize `<preferred-policy>` fall back to `<fallback-policy>`, while browsers that recognize both apply the **last (rightmost) valid token in the list**. The spec and MDN both document this ordering semantic explicitly — it is not an edge case, it's an intended, documented usage pattern.

`any(p in rp_lower for p in good_policies)` treats the header as an unordered bag of substrings. It has no concept of "which token is the effective one" and no concept of token position. Consequently, `<safe>, <unsafe>` and `<unsafe>, <safe>` are indistinguishable to this code, even though they produce opposite real-world behavior in every modern browser.

### Reproduction

```python
rp_lower = "strict-origin-when-cross-origin, unsafe-url"
good_policies = ["no-referrer", "strict-origin-when-cross-origin", "same-origin", "strict-origin"]
print(any(p in rp_lower for p in good_policies))  # -> True
```

In any current browser, the **effective** policy for this header value is `unsafe-url` (the last token, and it is universally recognized so nothing falls back further). `unsafe-url` sends the full referrer URL, including path and query string, cross-origin and over insecure transport — this is the single worst value the header can hold. The tool reports `severity: "low"`, `issue: "None"`.

### Impact

- **False negative on a real, spec-compliant configuration pattern.** This is not a malformed header being misparsed — it's a valid, documented header value being misjudged.
- Any downstream consumer (report, dashboard, compliance gate) trusts `severity: low` at face value and will not flag a site that is actually leaking full referrer URLs to third-party origins.
- Silent failure mode: no exception, no warning, just a wrong "clean" result.

### Proposed fix direction

Do not scan the whole raw string. Instead:

1. Split `rp` on `,`.
2. Strip whitespace from each resulting token and lowercase it.
3. Take the **last** token in the list — this is the effective policy per spec, assuming all listed tokens are recognized values (if you want to be fully rigorous, walk the list right-to-left and take the first token that is a recognized policy name, but at minimum stop scanning the whole string as an unordered bag).
4. Compare only that single resolved token against `good_policies` (exact match, not substring — see also the token-boundary concern raised in Issue 2, the same "substring vs. exact token" class of bug applies here too and is worth double-checking while this code is being touched).

### Test coverage required

Add cases to the header-analysis test file covering at minimum:

- `unsafe-url, strict-origin-when-cross-origin` → resolves to safe, `severity: low` (sanity check the fix doesn't just flip the answer unconditionally).
- `strict-origin-when-cross-origin, unsafe-url` → resolves to **unsafe**, must NOT be `severity: low`. This is the case that currently fails and is the core regression test for this issue.
- A single-value header (no comma) with a safe value, and a single-value header with an unsafe value — to confirm the non-list path still works after refactor.
- A list with trailing/leading whitespace around commas (`no-referrer , unsafe-url`) to confirm token-splitting is whitespace-tolerant.

Files to be touched :- `tools\headers_tool.py`, `tests\test_headers_tool.py`
Note :- Please comment before working on it

HeaTTap:
## Closes

Closes #192

## Summary

Parses comma-separated Referrer-Policy tokens and resolves the effective policy per spec ordering instead of checking substring matches across the raw header string.

## What changed

- Added module constants for valid and recommended referrer policies
- Split comma-separated tokens, stripped whitespace, and resolved the rightmost recognized policy
- Matched the resolved policy directly against recommended policies
- Added unit tests for fallback lists, single values, whitespace tolerance, and unknown token handling

## Notes

None

## Verification

- [x] All tests pass (`pytest`)

### DIFF
--- a/tests/test_headers_tool.py
+++ b/tests/test_headers_tool.py
@@ -401,6 +401,59 @@ class TestHeadersAnalyzer(unittest.TestCase):
         self.assertIn("may leak", rp["issue"])
         self.assertEqual(rp["severity"], "medium")
 
+    @patch("tools.headers_tool.requests.get")
+    def test_referrer_policy_fallback_list_safe_last(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Referrer-Policy": "unsafe-url, strict-origin-when-cross-origin",
+        })
+        result = headers_analyzer("example.com")
+        rp = result["headers"]["referrer-policy"]
+        self.assertTrue(rp["present"])
+        self.assertEqual(rp["issue"], "None")
+        self.assertEqual(rp["severity"], "low")
+
+    @patch("tools.headers_tool.requests.get")
+    def test_referrer_policy_fallback_list_unsafe_last(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Referrer-Policy": "strict-origin-when-cross-origin, unsafe-url",
+        })
+        result = headers_analyzer("example.com")
+        rp = result["headers"]["referrer-policy"]
+        self.assertTrue(rp["present"])
+        self.assertIn("may leak", rp["issue"])
+        self.assertEqual(rp["severity"], "medium")
+
+    @patch("tools.headers_tool.requests.get")
+    def test_referrer_policy_fallback_list_whitespace_tolerant(self, mock_get):
+        mock_get.return_value = _resp(200, {
+            "Referrer-Policy": "no-referrer ,  unsafe-url",
+        })
+        result = headers_analyzer("example.com")
+        rp = result["headers"]["referrer-policy"]
+        self.assertTrue(rp["present"])
+        self.assertIn("may leak", rp["issue"])
+        self.assertEqual(rp["severity"], "medium")
+
+        mock_get.return_value = _resp(200, {
+            "Referrer-Policy": "unsafe-url ,  no-referrer",
+        })
+        result2 = headers_analyzer("example.com")
+        rp2 = result2["headers"]["referrer-policy"]
+        self.assertTrue(rp2["present"])
+        self.assertEqual(rp2["issue"], "None")
+        self.assertEqual(rp2["severity"], "low")
+
+    @patch("tools.headers_tool.requests.get")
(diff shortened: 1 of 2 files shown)


gaoharimran29-glitch (approved):
Approved this correctly resolves #192.
The previous implementation treated Referrer-Policy as an unordered string and used substring matching, which could incorrectly classify a response as safe when its rightmost effective policy was unsafe. This change parses the comma-separated fallback list, resolves the rightmost recognized token, and evaluates that single effective policy using exact membership checks.
The regression coverage is comprehensive: it verifies both fallback-list orderings, whitespace normalization, and unknown-token handling. The change is focused, matches the header’s fallback semantics, and the test suite is passing. No changes requested , great work.

gaoharimran29-glitch:
@HeaTTap Its merged. If you're enjoying the project and would like to support its development, consider giving this repository a ⭐ on GitHub. It genuinely helps with visibility and encourages further contributions!

PR #205, opened 2026-09-11
fix(email-security): improve SPF, DKIM, and DMARC record validation (fixes #152)

### LINKED ISSUE #152: Improvement :-  SPF/DKIM/DMARC record validation + minor cleanup in email_security_check

AnandkumarMall:
## Closes

Closes #152

## Summary

Improves SPF, DKIM, and DMARC record validation accuracy and reliability per RFC specifications in `tools/email_security_tool.py`, preventing false positives, false negatives, and ambiguous configurations.

## What changed

- **SPF Improvements**:
  - Tokenized the SPF record to accurately parse the terminating `all` mechanism (`-all`, `~all`, `?all`, `+all`, bare `all`), preventing false matches inside unrelated substrings (e.g. `-allow`).
  - Added explicit `"missing"` status when a valid SPF record lacks an `all` mechanism, distinct from unparseable records (`"unknown"`).
  - Detected multiple SPF records: marks configuration invalid (`valid: False`), preserves all records, and provides actionable remediation guidance per RFC 7208.
- **DKIM Improvements**:
  - Implemented tag-aware DKIM validator confirming `v=DKIM1`, `p=`, and public key presence.
  - Distinguishes active keys (`found_selectors`), revoked/empty keys (`revoked_selectors`), malformed records, and unrelated TXT records containing `p=`.
- **DMARC Improvements**:
  - Detected multiple DMARC records: marks configuration ambiguous/invalid (`valid: False`), preserves all records, and recommends publishing a single DMARC record.
  - Improved `rua` tag validation to treat empty `rua=` as unconfigured and deduct score accordingly.
  - Added validation for `p=` values (`none`, `quarantine`, `reject`), flagging invalid values as `policy: "invalid"`.
- **Code Cleanups**:
  - Moved `import datetime` to top-level module imports.
  - Removed redundant `domain.lower()` in `email_security_check` in favor of `normalize_domain()`.
- **Test Coverage**:
  - Expanded `tests/test_email_security.py` covering multiple SPF records, multiple DMARC records, active vs revoked DKIM keys, empty `rua=`, invalid DMARC policy, and SPF tokenization edge cases.

## Verification

- [x] All tests pass (`uv run pytest tests/test_email_security.py -v` -> 12 passed; `uv run pytest` -> 552 passed)

### DIFF
--- a/tools/email_security_tool.py
+++ b/tools/email_security_tool.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 import concurrent.futures
+import datetime
 import dns.resolver
 import dns.exception
 from typing import Any, Dict, List, Tuple
@@ -45,7 +46,6 @@ def _discover_dynamic_selectors(domain: str) -> List[str]:
     dynamic_selectors = set()
     
     # 1. Inject temporal/historical pattern-matching common in enterprise setups (e.g., 2023, 2024, 2025, 2026)
-    import datetime
     current_year = datetime.datetime.now().year
     for year in range(current_year - 3, current_year + 1):
         dynamic_selectors.add(f"google{year}")
@@ -75,17 +75,62 @@ def _discover_dynamic_selectors(domain: str) -> List[str]:
     return list(dynamic_selectors)
 
 
+def _parse_dkim_record(record: str) -> Tuple[str, Dict[str, str]]:
+    """
+    Parse a TXT record for DKIM tags per RFC 6376.
+    Returns (status, tags) where status is one of:
+    - 'active': v=DKIM1 with non-empty p= key
+    - 'revoked': v=DKIM1 with empty p= key
+    - 'malformed': v= present but not DKIM1, or missing p= tag
+    - 'not_dkim': unrelated TXT record
+    """
+    parts = [p.strip() for p in record.split(";") if p.strip()]
+    tags: Dict[str, str] = {}
+    for part in parts:
+        if "=" in part:
+            k, v = part.split("=", 1)
+            tags[k.strip().lower()] = v.strip()
+
+    v_val = tags.get("v")
+    if v_val is not None:
+        if v_val.lower() != "dkim1":
+            return "malformed", tags
+        if "p" not in tags:
+            return "malformed", tags
+        if tags["p"]:
+            return "active", tags
+        return "revoked", tags
+
+    return "not_dkim", tags
+
+
 def _spf_policy(record: str) -> str:
-    record_lower = record.lower()
-    if "-all" in record_lower:
-        return "fail"
-    if "~all" in record_lower:
-        return "softfail"
-    if "?all" in record_lower:
-        return "neutral"
-    if "+all" in record_lower:
-        return "pass"
-    return "unknown"
+    """
+    Tokenize SPF record and locate the terminating 'all' mechanism.
+    Returns:
+    - 'fail' for -all
+    - 'softfail' for ~all
+    - 'neutral' for ?all
+    - 'pass' for +all or bare all
+    - 'missing' if record starts with v=spf1 but contains no 'all' mechanism
+    - 'unknown' for invalid/unparseable records
+    """
+    if not record.lower().startswith("v=spf1"):
+        return "unknown"
+
+    tokens = record.strip().split()
+    for token in reversed(tokens):
+        t_lower = token.lower()
+        if t_lower in ("-all", "~all", "?all", "+all", "all"):
+            if t_lower == "-all":
+                return "fail"
+            if t_lower == "~all":
+                return "softfail"
+            if t_lower == "?all":
+                return "neutral"
+            if t_lower in ("+all", "all"):
+                return "pass"
+    return "missing"
 
 
 def _check_spf(domain: str, recommendations: List[str]) -> Dict[str, Any]:
@@ -93,13 +138,26 @@ def _check_spf(domain: str, recommendations: List[str]) -> Dict[str, Any]:
 
     if failed:
         recommendations.append("Could not verify SPF — DNS lookup timed out.")
-        return {"found": False, "record": None, "policy": None, "score": 0}
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
 
     spf_records = [r for r in txt_records if r.lower().startswith("v=spf1")]
 
     if not spf_records:
         recommendations.append("SPF not found — add an SPF record.")
-        return {"found": False, "record": None, "policy": None, "score": 0}
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
+
+    if len(spf_records) > 1:
+        recommendations.append(
+            "Multiple SPF records found — SPF configuration is invalid. Consolidate the SPF mechanisms into a single SPF record."
+        )
+        return {
+            "found": True,
+            "valid": False,
+            "record": spf_records[0],
+            "records": spf_records,
+            "policy": None,
+            "score": 0,
+        }
 
     record = spf_records[0]
     policy = _spf_policy(record)
@@ -110,9 +168,25 @@ def _check_spf(domain: str, recommendations: List[str]) -> Dict[str, Any]:
         recommendations.append("SPF uses softfail (~all) — consider a hard fail (-all) for stronger protection")
     elif policy in ("neutral", "pass"):
         spf_score = 10
-        recommendations.append(f"SPF policy is '{policy}' — provides little protection.")
+        if policy == "pass":
+            recommendations.append("SPF policy is 'pass' (+all) — provides little protection.")
+        else:
+            recommendations.append(f"SPF policy is '{policy}' — provides little protection.")
+    elif policy == "missing":
+        spf_score = 15
+        recommendations.append("SPF record has no terminating 'all' mechanism — default policy is undefined.")
+    elif policy == "unknown":
+        spf_score = 0
+        recommendations.append("SPF record has an unrecognized or invalid 'all' mechanism.")
 
-    return {"found": True, "record": record, "policy": policy, "score": spf_score}
+    return {
+        "found": True,
+        "valid": True,
+        "record": record,
+        "records": spf_records,
+        "policy": policy,
+        "score": spf_score,
+    }
 
 
 def _check_dmarc(domain: str, recommendations: List[str]) -> Dict[str, Any]:
@@ -120,13 +194,26 @@ def _check_dmarc(domain: str, recommendations: List[str]) -> Dict[str, Any]:
 
     if failed:
         recommendations.append("Could not verify DMARC — DNS lookup timed out.")
-        return {"found": False, "record": None, "policy": None, "score": 0}
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
 
     dmarc_records = [r for r in txt_records if r.lower().startswith("v=dmarc1")]
 
     if not dmarc_records:
         recommendations.append("DMARC not found — add a DMARC record.")
-        return {"found": False, "record": None, "policy": None, "score": 0}
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
+
+    if len(dmarc_records) > 1:
+        recommendations.append(
+            "Multiple DMARC records found — DMARC configuration is ambiguous. Publish a single DMARC policy record."
+        )
+        return {
+            "found": True,
+            "valid": False,
+            "record": dmarc_records[0],
+            "records": dmarc_records,
+            "policy": None,
+            "score": 0,
+        }
 
     record = dmarc_records[0]
     tags = {}
@@ -136,7 +223,20 @@ def _check_dmarc(domain: str, recommendations: List[str]) -> Dict[str, Any]:
             k, v = part.split("=", 1)
             tags[k.strip().lower()] = v.strip()
 
-    policy = tags.get("p", "none").lower()
+    raw_policy = tags.get("p")
+    valid = True
+    if raw_policy is None:
+        policy = "none"
+    else:
+        policy_lower = raw_policy.lower()
+        if policy_lower in ("reject", "quarantine", "none"):
+            policy = policy_lower
+        else:
+            policy = "invalid"
+            valid = False
+            recommendations.append(
+                f"DMARC policy 'p={raw_policy}' is invalid — must be 'none', 'quarantine', or 'reject'."
+            )
 
     dmarc_score = 10
     if policy == "reject":
@@ -148,8 +248,18 @@ def _check_dmarc(domain: str, recommendations: List[str]) -> Dict[str, Any]:
     if "rua" not in tags:
         dmarc_score = max(0, dmarc_score - 5)
         recommendations.append("DMARC has no rua= reporting address.")
+    elif not tags["rua"]:
+        dmarc_score = max(0, dmarc_score - 5)
+        recommendations.append("DMARC has an empty rua= tag with no reporting address.")
 
-    return {"found": True, "record": record, "policy": policy, "score": dmarc_score}
+    return {
+        "found": True,
+        "valid": valid,
+        "record": record,
+        "records": dmarc_records,
+        "policy": policy,
+        "score": dmarc_score,
+    }
 
 
 def _check_dkim(domain: str, recommendations: List[str]) -> Dict[str, Any]:
@@ -157,28 +267,40 @@ def _check_dkim(domain: str, recommendations: List[str]) -> Dict[str, Any]:
     dynamic_keys = _discover_dynamic_selectors(domain)
     selectors_to_check = list(set(BASE_DKIM_SELECTORS + dynamic_keys))
 
-    def check_selector(selector: str):
+    def check_selector(selector: str) -> Tuple[str, str | None]:
         records, _failed = _query_txt(f"{selector}._domainkey.{domain}")
-        if any("v=dkim1" in r.lower() or "p=" in r.lower() for r in records):
-            return selector
-        return None
+        for r in records:
+            status, _tags = _parse_dkim_record(r)
+            if status == "active":
+                return selector, "active"
+            elif status == "revoked":
+                return selector, "revoked"
+        return selector, None
 
     # Threading prevents the added dynamic keys from slowing down execution time
     with concurrent.futures.ThreadPoolExecutor(max_workers=len(selectors_to_check)) as executor:
         results = executor.map(check_selector, selectors_to_check)
-        found_selectors = [s for s in results if s is not None]
+        results_list = list(results)
+        found_selectors = [s for s, status in results_list if status == "active"]
+        revoked_selectors = [s for s, status in results_list if status == "revoked"]
 
     found = len(found_selectors) > 0
     dkim_score = 35 if found else 0
 
     if not found:
-        recommendations.append("DKIM not found — add DKIM record to prevent spoofing")
+        if revoked_selectors:
+            recommendations.append(
+                f"DKIM key(s) revoked for selector(s): {', '.join(revoked_selectors)} — configure an active public key."
+            )
+        else:
+            recommendations.append("DKIM not found — add DKIM record to prevent spoofing")
 
     return {
-        "found": found, 
+        "found": found,
         "selectors_checked": selectors_to_check,
         "found_selectors": found_selectors,
-        "score": dkim_score
+        "revoked_selectors": revoked_selectors,
+        "score": dkim_score,
     }
 
 
@@ -191,8 +313,6 @@ def email_security_check(domain: str) -> dict:
     selectors used by major email providers, since DKIM selectors
     cannot be discovered via DNS without already knowing them.
     """
-    domain = domain.strip().lower()
-
     domain = normalize_domain(domain)
     if not is_valid_domain(domain):
         return {"success": False, "error": "Invalid domain format"}
@@ -227,4 +347,4 @@ def email_security_check(domain: str) -> dict:
         }
 
     except Exception as e:
-        return {"success": False, "error": str(e)}
+        return {"success": False, "error": str(e)}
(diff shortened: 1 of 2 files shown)


tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    if v_val is not None:
+        if v_val.lower() != "dkim1":
+            return "malformed", tags
+        if "p" not in tags:
+            return "malformed", tags
+        if tags["p"]:
+            return "active", tags
+        return "revoked", tags
+
+    return "not_dkim", tags
```
gaoharimran29-glitch: RFC 6376 Section 3.6.1: The v tag is optional but recommended. If it is omitted, the version defaults to DKIM1. If a record lacks a v tag but contains a valid p tag, it is a valid active record. Your current logic flags these legitimate records as `not_dkim`.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    parts = [p.strip() for p in record.split(";") if p.strip()]
+    tags: Dict[str, str] = {}
+    for part in parts:
+        if "=" in part:
+            k, v = part.split("=", 1)
+            tags[k.strip().lower()] = v.strip()
+
+    v_val = tags.get("v")
+    if v_val is not None:
+        if v_val.lower() != "dkim1":
```
gaoharimran29-glitch: As the comment in line 80 say the parser is following RFC 6376. But RFC 6376 Section 3.2 : The version tag v is explicitly defined as case-sensitive and must be exactly lowercase `DKIM1`. An email signature containing `v=dkim1` or `v=DKIM1`  is invalid. Your function would mistakenly accept a lowercase or mixed-case version tag as valid, which breaks compliance.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    - 'revoked': v=DKIM1 with empty p= key
+    - 'malformed': v= present but not DKIM1, or missing p= tag
+    - 'not_dkim': unrelated TXT record
+    """
+    parts = [p.strip() for p in record.split(";") if p.strip()]
+    tags: Dict[str, str] = {}
+    for part in parts:
+        if "=" in part:
+            k, v = part.split("=", 1)
+            tags[k.strip().lower()] = v.strip()
```
gaoharimran29-glitch: RFC 6376 Section 3.2: Tag names are strictly defined as alpha-num sequences without internal or surrounding whitespace. A record formatted as `v = DKIM1` is structurally malformed because of the space before the equal sign. Stripping the key hides structural defects that should cause the record to fail validation. Also Lowercases tag names, so invalid `P=key` is treated as valid `p=key`; DKIM tag names are case-sensitive. It also silently overwrites duplicate tags, although duplicates invalidate the entire tag list.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    Parse a TXT record for DKIM tags per RFC 6376.
+    Returns (status, tags) where status is one of:
+    - 'active': v=DKIM1 with non-empty p= key
+    - 'revoked': v=DKIM1 with empty p= key
+    - 'malformed': v= present but not DKIM1, or missing p= tag
+    - 'not_dkim': unrelated TXT record
+    """
+    parts = [p.strip() for p in record.split(";") if p.strip()]
+    tags: Dict[str, str] = {}
+    for part in parts:
```
gaoharimran29-glitch, line 114: Accepts `p=key`; `v=DKIM1`, though `v=` must be the first tag when present. according to RFC 6376

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+                dynamic_selectors.update(["selector1", "msft", "microsoft"])
+    except Exception:
+        pass  # Fail gracefully if MX discovery hits a hitch
+
+    return list(dynamic_selectors)
+
+
+def _parse_dkim_record(record: str) -> Tuple[str, Dict[str, str]]:
+    """
+    Parse a TXT record for DKIM tags per RFC 6376.
```
gaoharimran29-glitch: Suggestion :- Add a link for reference for RFC 6376 .

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    - 'neutral' for ?all
+    - 'pass' for +all or bare all
+    - 'missing' if record starts with v=spf1 but contains no 'all' mechanism
+    - 'unknown' for invalid/unparseable records
+    """
+    if not record.lower().startswith("v=spf1"):
+        return "unknown"
+
+    tokens = record.strip().split()
+    for token in reversed(tokens):
```
gaoharimran29-glitch: Your function loops through the tokens backward (`reversed(tokens)`) and returns immediately on the first all mechanism it finds. RFC 7208 Section 4.6.2: Evaluates mechanisms sequentially from left to right.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+        if t_lower in ("-all", "~all", "?all", "+all", "all"):
+            if t_lower == "-all":
+                return "fail"
+            if t_lower == "~all":
+                return "softfail"
+            if t_lower == "?all":
+                return "neutral"
+            if t_lower in ("+all", "all"):
+                return "pass"
+    return "missing"
```
gaoharimran29-glitch, line 214: If a record doesn't end in all, your code returns "missing" . RFC 7208 Section 6.1: An SPF record does not require an all mechanism if it includes a redirect modifier (e.g., `v=spf1 redirect=_://google.com`). The target domain's all policy is inherited instead.

tools/email_security_tool.py
gaoharimran29-glitch: Your function loops through the tokens backward (`reversed(tokens)`) and returns immediately on the first all mechanism it finds. RFC 7208 Section 4.6.2: Evaluates mechanisms sequentially from left to right.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    Returns:
+    - 'fail' for -all
+    - 'softfail' for ~all
+    - 'neutral' for ?all
+    - 'pass' for +all or bare all
+    - 'missing' if record starts with v=spf1 but contains no 'all' mechanism
+    - 'unknown' for invalid/unparseable records
+    """
+    if not record.lower().startswith("v=spf1"):
+        return "unknown"
```
gaoharimran29-glitch: RFC 7208 Section 4.1: The version identifier must be lowercase (`v=spf1`). A record beginning with `V=SPF1` is structurally invalid and must be completely ignored by the receiving server. Your function incorrectly accepts mixed-case versions. Also `startswith("v=spf1")` still accepts malformed values such as `v=spf1foo` or `v=spf1`; `-all` as SPF records. That collapses an unparseable record into missing/a policy result, contrary to the issue's requirement that malformed/unknown records remain distinct from a valid SPF record with no all mechanism. Tokenize first, require the first token to equal `v=spf1`, apply the same predicate when collecting SPF records, and add regression tests.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
+
+    if len(dmarc_records) > 1:
+        recommendations.append(
+            "Multiple DMARC records found — DMARC configuration is ambiguous. Publish a single DMARC policy record."
+        )
+        return {
+            "found": True,
+            "valid": False,
+            "record": dmarc_records[0],
```
gaoharimran29-glitch: Same issue as in line 156

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+        return {"found": False, "valid": False, "record": None, "records": [], "policy": None, "score": 0}
+
+    if len(spf_records) > 1:
+        recommendations.append(
+            "Multiple SPF records found — SPF configuration is invalid. Consolidate the SPF mechanisms into a single SPF record."
+        )
+        return {
+            "found": True,
+            "valid": False,
+            "record": spf_records[0],
```
gaoharimran29-glitch: When multiple SPF records are found, `record: spf_records[0]` still silently selects the first record. Since the requirement says not to choose one, Make it to return `record: None` and preserve the candidates only in `records` This would make the invalid/ambiguous state explicit.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+    for part in record.split(";"):
+        part = part.strip()
+        if "=" in part:
+            k, v = part.split("=", 1)
+            tags[k.strip().lower()] = v.strip()
+
+    raw_policy = tags.get("p")
+    valid = True
+    if raw_policy is None:
+        policy = "none"
```
gaoharimran29-glitch, line 325: RFC 9989 Section 4.7 (Current Standard): Your fallback logic matches the standard. The spec dictates: "If this tag is not present in an otherwise syntactically valid DMARC Policy Record, then the record is treated as if it included 'p=none'" . But adding a recommendation will be helpful here instead of labeling it as perfectly valid
```suggestion
        policy = "none"
        recommendations.append(
        "DMARC record is missing the 'p=' tag. While the new RFC 9989 standard defaults "
        "this to 'none', legacy mail systems may ignore this record entirely."
    )
```

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+
+    def check_selector(selector: str) -> Tuple[str, str | None]:
+        records, _failed = _query_txt(f"{selector}._domainkey.{domain}")
+        for r in records:
+            status, _tags = _parse_dkim_record(r)
+            if status == "active":
+                return selector, "active"
+            elif status == "revoked":
+                return selector, "revoked"
+        return selector, None
```
gaoharimran29-glitch: `_parse_dkim_record()` can return malformed, but this branch only preserves active and revoked; malformed selectors fall through to `None` and become indistinguishable from selectors with no DKIM record. Issue #152 requires the scanner to distinguish active keys, revoked/empty keys, malformed records, and no DKIM record. Collect malformed selectors into a malformed_selectors field (or equivalent), return it in the result, and add assertions for the `existing v=DKIM1-without-p=` test case.

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+        spf_score = 20
+        recommendations.append("SPF uses softfail (~all) — consider a hard fail (-all) for stronger protection")
+    elif policy in ("neutral", "pass"):
+        spf_score = 10
+        if policy == "pass":
+            recommendations.append("SPF policy is 'pass' (+all) — provides little protection.")
+        else:
+            recommendations.append(f"SPF policy is '{policy}' — provides little protection.")
+    elif policy == "missing":
+        spf_score = 15
```
gaoharimran29-glitch: A record with a `"missing"` or completely undefined all mechanism defaults to an implicit neutral or undefined state in real-world mail servers. However, practically, a missing terminator is a major structural omission or syntax oversight by a domain administrator. A broken/incomplete record (15) should not receive a higher health score than a completed, fully syntactically valid record set to "neutral" (10).

tools/email_security_tool.py
```diff
@@ -1,230 +1,350 @@
...
+            tags[k.strip().lower()] = v.strip()
+
+    v_val = tags.get("v")
+    if v_val is not None:
+        if v_val.lower() != "dkim1":
+            return "malformed", tags
+        if "p" not in tags:
+            return "malformed", tags
+        if tags["p"]:
+            return "active", tags
```
gaoharimran29-glitch: This only verifies that `p=` is non-empty, so arbitrary text such as `v=DKIM1; p=not-a-public-key` is classified as an active key. Issue #152 requires `p=` to contain a non-empty public key. Please validate the normalized value as base64 public-key data, classify invalid values as malformed, and add a regression test. A full cryptographic-key parser is unnecessary here.

gaoharimran29-glitch (changes requested):
Thanks for the focused implementation and accompanying regression coverage. The PR addresses the central SPF, DKIM, DMARC, and cleanup goals in #152 well.  

But there are few issues in this code for which i requested changes please see inline comments below and fix them.

A few broader RFC inconsistencies also exist in pre-existing project code. Those are acknowledged but are not being treated as blockers for this PR unless they are introduced or directly affected here; we will track the broader standards-alignment work in a follow-up issue.

tools/email_security_tool.py
```diff
@@ -75,51 +75,127 @@ def _discover_dynamic_selectors(domain: str) -> List[str]:
...
+
+    for part in parts:
+        if "=" not in part:
+            return "malformed", tags
+
+        raw_k, raw_v = part.split("=", 1)
+
+        # RFC 6376 §3.2: Tag names are strictly defined as alpha-num sequences
+        # without internal or surrounding whitespace. A space before '=' is invalid.
+        if not raw_k.isalnum():
```
gaoharimran29-glitch, line 136: `isalnum()` does not match the RFC 6376 tag-name rule. It rejects valid extension tags containing _ (for example, `x_foo=value`), causing an otherwise valid DKIM key to be marked malformed. It also accepts invalid names beginning with a digit and non-ASCII letters. Please validate tag names with an ASCII pattern: first character [A-Za-z], followed by zero or more [A-Za-z0-9_]; unknown valid tags should be ignored.

gaoharimran29-glitch (changes requested):
Thanks for addressing the requested changes—I appreciate the effort. There is just one remaining item; please see the inline comment. Once that is resolved, this PR will be ready to merge.

gaoharimran29-glitch:
Just to clarify what happened here: this PR was merged inadvertently while I was preparing to merge another approved implementation for the same issue. I’ve since reverted the merge so that we can proceed with the other implementation and keep a single solution for this issue in the codebase.

I want to emphasize that this decision is **not a reflection of the quality of your implementation or your capabilities as a developer**. Your PR was reviewed and was a valid approach to solving the issue. The fact that it won’t remain merged is purely due to the situation around having two independent implementations for the same issue, and the decision to proceed with the alternate one.

Apologies for the confusion caused by the accidental merge, and thank you for taking the time to investigate and contribute a solution. I really appreciate the effort that went into this PR.

Issue #169, opened 2026-08-20
ci(workflow): add Ruff linting to CI and standardize PR checks


gaoharimran29-glitch:
## Context

The repository currently has a `.github/workflows/tests.yml` workflow that executes the test suite using:  It Depends on #167 and to be take only after #167 gets fixed

```bash
pytest tests/ -v
```

Pull requests are not currently validated against the repository's Ruff linting and formatting requirements. This allows PRs to pass the test suite while still introducing lint violations or unformatted code.
The repository should enforce the following checks for every pull request:

```bash
uv run pytest tests/ -v
uv run ruff check .
uv run ruff format --check .
```

The existing test command should be updated from `pytest tests/ -v` to `uv run pytest tests/ -v` so that CI and local contributor workflows use the project's `uv` environment consistently.

### 1. Extend `tests.yml`

Update `.github/workflows/tests.yml` to execute the following checks:

```bash
uv run pytest tests/ -v
uv run ruff check .
uv run ruff format --check .
```

Replace the existing:

```bash
pytest tests/ -v
```

with:

```bash
uv run pytest tests/ -v
```

The implementation should use the repository's existing `uv` setup and dependency configuration rather than introducing a separate dependency installation mechanism.
The workflow should return a non-zero exit status when any check fails so that GitHub marks the PR checks as failed.
If practical, keep the test, lint, and formatting checks as separate CI steps to make failures immediately identifiable from the GitHub Actions UI.

### 2. Update `CONTRIBUTING.md`

Update all relevant sections of `CONTRIBUTING.md` to reflect the new PR validation workflow.
Wherever the existing test command is documented as:

```bash
pytest tests/ -v
```

update it to:

```bash
uv run pytest tests/ -v
```

Add the Ruff validation commands where appropriate:

```bash
uv run ruff check .
uv run ruff format --check .
```

Ensure the contributor documentation consistently instructs contributors to run the same checks that are enforced by CI before opening a pull request.
Review all relevant references to testing, linting, formatting, pull-request validation, and development setup rather than updating only a single occurrence.
Do not leave conflicting instructions that continue to recommend running `pytest tests/ -v` directly.

### 3. Update `.github/PULL_REQUEST_TEMPLATE.md`

Update the `## Verification` section in `.github/PULL_REQUEST_TEMPLATE.md`.
Replace:

```markdown
- [ ] All tests pass (`<test command here>`)
```

with:

```markdown
[...]

PR #206, opened 2026-09-11
feat(dns): add DNSSEC records, reverse PTR lookups, and CNAME chain tracing

### LINKED ISSUE #144: Enhancement: Improve DNS Enumeration Coverage, Performance, and Error Handling

AnandkumarMall:
## Closes

Closes #144

## Summary

Completes the remaining items for DNS Enumeration:
1. Enumeration and structured formatting of DNSSEC records (`DNSKEY`, `DS`, `RRSIG`, and `NSEC`).
2. Recursive CNAME chain resolution with loop detection and max hop limit.
3. Reverse DNS PTR lookups for resolved IPv4 (`A`) and IPv6 (`AAAA`) addresses.

## What changed

- **`tools/dns_tool.py`**:
  - Added DNSSEC record query support using `_DnsExecutor` for `DNSKEY`, `DS`, `RRSIG`, and `NSEC`.
  - Added specialized DNSSEC formatters (`_format_dnskey_record`, `_format_ds_record`, `_format_rrsig_record`, `_format_nsec_record`) with raw string fallbacks.
  - Added `_resolve_cname_chain` to follow aliases up to depth 5 while guarding against alias loops and recursion cycles.
  - Added `_resolve_ptr_records` leveraging `dns.reversename` to resolve reverse PTR records for all discovered A/AAAA IPs.
  - Preserved backward compatibility for the top-level `records` dictionary (8 core types) and returned `dnssec_records`, `dnssec_errors`, `cname_chain`, `ptr_records`, and `ptr_errors` as top-level keys.
- **`tests/test_dns_enumeration.py`**:
  - Added comprehensive unit tests for DNSSEC formatting, CNAME chain resolution & loop termination, and reverse DNS PTR resolution.
  - Ensured response shape test accounts for the new response keys.
- **Documentation**:
  - Updated `README.md` tool inventory table for `dns_enumeration`.
  - Updated `docs/tools/dns_enumeration.md` with example JSON output and field reference tables for DNSSEC, PTR, and CNAME chains.

## Notes

- Preserving the exact 8-key structure of `records` prevents breaking existing signal extractors and downstream tests that assert `set(result["records"]) == {"A", "AAAA", "MX", "NS", "TXT", "CNAME", "SOA", "CAA"}`.

## Verification


gaoharimran29-glitch:
@SemTiOne Passing this your way! Please review whenever you get a free slot. I’m a bit tied up with college and internship at the moment. Thanks a lot!

gaoharimran29-glitch:
@AnandkumarMall please fix the conflicts so @SemTiOne can review it soon.

AnandkumarMall:
@gaoharimran29-glitch @SemTiOne Conflicts have been resolved and all DNS tests pass! Ready for review.

tools/dns_tool.py
```diff
@@ -26,10 +29,14 @@ def __exit__(self, exc_type, exc_value, traceback):
 RESOLVER_LIFETIME = 5
 SUBDOMAIN_LIFETIME = 3
 MAX_CONCURRENT_LOOKUPS = 10
+MAX_CNAME_DEPTH = 5
 
 # Record types enumerated for the target domain.
 RECORD_TYPES = ["A", "AAAA", "MX", "NS", "TXT", "CNAME", "SOA", "CAA"]
 
+# DNSSEC record types enumerated for the target domain (issue #144, item 4).
+DNSSEC_RECORD_TYPES = ["DNSKEY", "DS", "RRSIG", "NSEC"]
```
gaoharimran29-glitch: `[BLOCKING]` Issue spec is literally `NSEC / NSEC3` (where applicable) and this only ships `NSEC`. `NSEC3` is the more commonly deployed denial-of-existence scheme in the wild (opt-out zones basically require it), so silently dropping it undercuts the whole point of this item — a scan against an `NSEC3-only` zone will report "no `NSEC-family` records" and give a false negative on `DNSSEC` posture. Either wire up `NSEC3` parsing too or drop a comment here explaining why it's out of scope for this PR, don't just omit it quietly.

tools/dns_tool.py
```diff
@@ -164,6 +171,107 @@ def _format_caa_record(record) -> dict:
...
+        "key_tag": getattr(record, "key_tag", None),
+        "signer": _clean_name(getattr(record, "signer", "")),
+    }
+
+
+def _format_nsec_record(record) -> dict:
+    next_name = getattr(record, "next", None)
+    return {
+        "next": _clean_name(next_name) if next_name else str(record),
+    }
```
gaoharimran29-glitch, line 280: `[BLOCKING]` This only pulls .next. The actual recon value of an `NSEC` record is the type bitmap (`record.windows` in `dnspython`) — it tells you what other record types exist at that name, which is exactly the kind of `"security-focused reconnaissance"` the issue calls out. Right now we're throwing away the interesting half of the record.

tools/dns_tool.py
```diff
@@ -164,6 +171,107 @@ def _format_caa_record(record) -> dict:
...
+            if next_target.lower() in seen:
+                break
+            chain.append(next_target)
+            seen.add(next_target.lower())
+            current = next_target
+        except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN, dns.exception.DNSException):
+            break
+        except Exception:
+            break
+    return chain
```
gaoharimran29-glitch: `[BLOCKING]` `DNSException` already swallows basically every `dnspython` failure mode (`Timeout`, `NoNameservers`, you name it), so that trailing bare `except Exception`: break only fires for genuinely unexpected bugs — and it eats them with zero trace. Every other lookup helper in this file (`_resolve_ptr_records`, `_lookup_subdomain`, the new `dnssec` block) returns an errors dict specifically so we don't do this. This function should follow the same convention — return (chain, errors) — otherwise a chain that got cut short by a real bug looks identical to a chain that terminated normally on an A record, and that's exactly the kind of failure-masking issue #1 was supposed to kill.

tools/dns_tool.py
```diff
@@ -164,6 +171,107 @@ def _format_caa_record(record) -> dict:
...
+    current = str(initial_cnames[0]).rstrip(".")
+    chain.append(current)
+    seen.add(current.lower())
+
+    for _ in range(MAX_CNAME_DEPTH):
+        try:
+            answers = resolver.resolve(current, "CNAME", lifetime=RESOLVER_LIFETIME)
+            next_target = _clean_name(answers[0])
+            if next_target.lower() in seen:
+                break
```
gaoharimran29-glitch, line 302: `[NON-BLOCKING]` When a cycle is detected we just stop and return the partial chain — there's no signal in the output that a loop was hit vs. normal termination. For a recon tool a `CNAME` loop is itself a finding (misconfig or worse), not just an edge case to swallow. Worth surfacing something like a `cycle_detected`/ `truncated` flag on the result rather than making the caller infer it.

tools/dns_tool.py
```diff
@@ -164,6 +171,107 @@ def _format_caa_record(record) -> dict:
...
+            ptr_records[str(ip)] = [_clean_name(r.target) for r in answers]
+        except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN):
+            ptr_records[str(ip)] = []
+        except LOOKUP_ERRORS as exc:
+            ptr_records[str(ip)] = []
+            ptr_errors[str(ip)] = type(exc).__name__
+        except Exception as exc:
+            ptr_records[str(ip)] = []
+            ptr_errors[str(ip)] = f"unexpected: {type(exc).__name__}"
+    return ptr_records, ptr_errors
```
gaoharimran29-glitch: `[NON-BLOCKING]` This runs as a plain sequential for loop even though it executes inside the `_DnsExecutor` with block that issue #2 set up specifically so independent DNS lookups don't serialize. Every other batch of independent lookups in this function (record types, `SRV`, subdomains, `DNSSEC`) goes through `executor.submit`. PTR lookups are just as independent per-IP as those — for a domain with a handful of `A/AAAA` records this reintroduces the exact latency problem #2 was closed to fix.

tools/dns_tool.py
gaoharimran29-glitch, line 413: `[BLOCKING]` Docstring still describes the pre-#4/#7/#8 return shape. `cname_chain`, `dnssec_records`/`dnssec_errors`, `ptr_records`/`ptr_errors` are all missing from it. Anyone reading this function signature has no idea these fields exist.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +951,133 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
+        self.assertEqual(len(dnssec["DS"]), 1)
+        self.assertEqual(dnssec["DS"][0]["key_tag"], 12345)
+        self.assertEqual(dnssec["DS"][0]["digest"], "abcdef123456")
+
+        self.assertEqual(len(dnssec["RRSIG"]), 1)
+        self.assertEqual(dnssec["RRSIG"][0]["type_covered"], "A")
+        self.assertEqual(dnssec["RRSIG"][0]["signer"], "example.com")
+
+        self.assertEqual(len(dnssec["NSEC"]), 1)
+        self.assertEqual(dnssec["NSEC"][0]["next"], "next.example.com")
```
gaoharimran29-glitch, line 1027: `[BLOCKING]` This only covers the happy path. There's no test exercising `dnssec_errors` — e.g. what happens when a `DNSKEY` lookup throws `Timeout` or `NoNameservers` instead of `NoAnswer`. Given that "don't mask unexpected failures" is the whole premise of issue #1, the new `DNSSEC` path needs a test asserting that a genuine lookup error lands in `dnssec_errors[type]` with the right exception name, not just that a clean 4-for-4 success case formats correctly.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +951,133 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
         self.assertEqual(result["resolver"]["nameservers"], ["1.1.1.1", "8.8.8.8"])
-        self.assertEqual(result["resolver"]["lifetime"], 5)
+    @patch("tools.dns_tool.dns.resolver.Resolver")
+    def test_dnssec_records_enumerated_and_formatted(self, mock_resolver_class):
+        import dns.resolver as real_dns
+
+        resolver = Mock()
+        mock_resolver_class.return_value = resolver
+
+        def side_effect(domain, rtype, lifetime=5, tcp=False):
```
gaoharimran29-glitch: `[NON-BLOCKING]` `lifetime=5` is hardcoded into the mock signature instead of referencing `RESOLVER_LIFETIME` from the module. If that constant ever changes, this test keeps passing for the wrong reason — it's not actually asserting anything about the real default, just matching itself.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +951,133 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
+                return self._make_resolver_answer(["cdn.provider.net."])
+            if domain == "cdn.provider.net" and rtype == "CNAME":
+                raise real_dns.NoAnswer
+            raise real_dns.NoAnswer
+
+        resolver.resolve.side_effect = side_effect
+        result = dns_enumeration("example.com")
+
+        self.assertTrue(result["success"])
+        self.assertEqual(result["cname_chain"], ["alias.example.net", "cdn.provider.net"])
```
gaoharimran29-glitch: `[NON-BLOCKING]` Only tests a clean 2-hop chain ending in `NoAnswer`. Given `MAX_CNAME_DEPTH = 5` exists specifically as a safety cap, there should be a test that actually drives a non-cyclic chain past that depth (5 distinct hops, no repeats) to confirm the cap kicks in and the function returns cleanly instead of just trusting the `range(MAX_CNAME_DEPTH)` loop bound by inspection. That's a real DoS vector for a recon tool pointed at attacker-controlled infra — worth pinning down with a test, not just code review.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +951,133 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
+                return self._make_resolver_answer(["hop1.com."])
+            if domain == "hop1.com" and rtype == "CNAME":
+                return self._make_resolver_answer(["loop.com."])
+            raise real_dns.NoAnswer
+
+        resolver.resolve.side_effect = side_effect
+        result = dns_enumeration("loop.com")
+
+        self.assertTrue(result["success"])
+        self.assertEqual(result["cname_chain"], ["hop1.com"])
```
gaoharimran29-glitch, line 1106: `[NON-BLOCKING]` This confirms the chain stops on a cycle but doesn't assert anything distinguishing "stopped because cycle" from "stopped because `NoAnswer`" — ties back to the missing-signal comment on `_resolve_cname_chain` above. Once that function starts returning error/status info, this test should assert on it.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +951,133 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
+            raise real_dns.NoAnswer
+
+        resolver.resolve.side_effect = side_effect
+        result = dns_enumeration("example.com")
+
+        self.assertTrue(result["success"])
+        self.assertIn("192.0.2.1", result["ptr_records"])
+        self.assertEqual(result["ptr_records"]["192.0.2.1"], ["host1.example.com"])
+        self.assertEqual(result["ptr_errors"], {})
+
```
gaoharimran29-glitch, line 1157: `[BLOCKING]` Only tests the success path for a single IP. There's no test for `ptr_errors` getting populated on an unexpected failure (mirrors the `DNSSEC` gap above), and no test for the dedup logic (`if ip not in ip_addresses`) — e.g. the same IP showing up in both the `A` and `AAAA` answer sets. That dedup branch is new code with zero coverage right now; if it regresses, nothing here would catch it.

gaoharimran29-glitch (changes requested):
## Overall Review

Nice work on this one — #4/#7/#8 were the trickier back half of the tracking issue (DNSSEC formatting, PTR reverse-mapping, CNAME chain walking with cycle detection) and the core plumbing for all three is solid. The way you slotted DNSSEC and PTR into the existing `_DnsExecutor` pattern shows you actually read the prior PRs instead of bolting stuff on sideways, and the cycle-detection logic in the CNAME walker is correct in the cases you tested. Appreciate the effort here, this isn't a small diff.

That said, I can't approve as-is — going **request changes**, see inline comments. 
Fix them and then the pr will be ready to merge.

One more thing, separate from the above: **ruff is flagging a handful of issues in this diff.** Heads up if you're not already running it locally — PR #212 got the whole repo onto `uv run ruff format --check`, so it's actually enforced now even though `contributing.md` hasn't caught up to mention it yet (my bad, that's on us , issue is already listed for it #169). Can you run `uv run ruff format --check` on your branch and clean up what it flags before the next round of review?

Solid foundation overall, just needs another pass on the error-handling consistency and test coverage before this is mergeable. Ping me once you've pushed updates and I'll take another look.

AnandkumarMall:
Hi @gaoharimran29-glitch,

Thank you for the thorough review and helpful feedback! I have addressed all the review comments in the latest push:

1. **NSEC3 & Type Bitmap Support**:
   - Added NSEC3 to DNSSEC_RECORD_TYPES with complete parsing (lgorithm, lags, iterations, salt, base32 
ext hash, and 	ypes).
   - Extracted 	ypes (the type bitmap) from windows using dnspython.rdtypes.util.Bitmap for both NSEC and NSEC3 records.
2. **CNAME Chain Error Handling & Loop Signaling**:
   - Refactored _resolve_cname_chain to follow the standard (chain, errors) convention without swallowing unexpected exceptions.
   - Surfaced cycle_detected: True when a loop is detected and 	runcated: True when hitting MAX_CNAME_DEPTH.
   - Added cname_errors to the returned dictionary.
3. **Parallel PTR Resolution**:
   - PTR reverse DNS lookups are now dispatched concurrently across the thread pool via executor.submit.
4. **Docstrings & Documentation**:
   - Updated dns_enumeration docstrings and docs to describe all fields (cname_chain, cname_errors, dnssec_records/errors, ptr_records/errors).
5. **Test Coverage & Formatting**:
   - Added tests for dnssec_errors on lookup failure (NoNameservers, Timeout).
   - Added tests for ptr_errors and IP deduplication between A and AAAA.
   - Added test for CNAME truncation when depth exceeds MAX_CNAME_DEPTH.
   - Replaced hardcoded lifetime=5 with RESOLVER_LIFETIME across tests.
   - Ran uv run ruff format and uv run ruff check (all clean), and verified all 608 tests pass.

Ready for your review!

tools/dns_tool.py
```diff
@@ -164,6 +171,191 @@ def _format_caa_record(record) -> dict:
...
+        try:
+            next_hash = record._next_text()
+        except Exception:
+            next_hash = ""
+    if not next_hash:
+        next_raw = getattr(record, "next", None)
+        if isinstance(next_raw, bytes):
+            next_hash = base64.b32encode(next_raw).decode("ascii").rstrip("=")
+        elif next_raw is not None:
+            next_hash = str(next_raw)
```
gaoharimran29-glitch: `[NON-BLOCKING]` Dug into dnspython's actual `NSEC3.to_text()` to check this — `_next_text()` is real, and it does `base64.b32encode(self.next).translate(b32_normal_to_hex).lower()`, i.e. `base32hex`, which is what `RFC 5155` actually requires for the hashed owner name. Good that you got the encoding right in the primary path. But the fallback — `base64.b32encode(next_raw).decode("ascii").rstrip("=")` — uses standard `base32`, no `.translate(b32_normal_to_hex)`. If `_next_text` is ever missing (mocked object, future `dnspython` version, whatever), this silently produces a hash string in the wrong alphabet — inconsistent with the primary path's output and with what dig/zone files show for the same record. Since you're leaning on a private (prefixed) method that isn't part of `dnspython's` public API contract, this fallback isn't just defensive padding — it's the path that runs the day `dnspython` renames or drops that method, and right now it's wrong when it runs. Either apply the same translate step in the fallback, or inline the full `base32hex` logic yourself instead of depending on `_next_text` at all.

tests/test_dns_enumeration.py
```diff
@@ -939,7 +953,258 @@ def test_resolver_metadata_in_output(self, mock_resolver_class):
...
         self.assertIn("resolver", result)
         self.assertEqual(result["resolver"]["nameservers"], ["1.1.1.1", "8.8.8.8"])
-        self.assertEqual(result["resolver"]["lifetime"], 5)
+
+    @patch("tools.dns_tool.dns.resolver.Resolver")
+    def test_dnssec_records_enumerated_and_formatted(self, mock_resolver_class):
+        import dns.resolver as real_dns
+
+        resolver = Mock()
+        mock_resolver_class.return_value = resolver
```
gaoharimran29-glitch: `[NON-BLOCKING]` Related to the previous comment — `test_dnssec_records_enumerated_and_formatted` mocks `rec._next_text` directly, so it only exercises the primary path. The fallback branch (no `_next_text`, raw bytes on `.next`) has zero coverage, which is exactly how the alphabet mismatch above went unnoticed. A quick test with a plain object that has .next as raw bytes but no `_next_text` attr would pin down expected output and would've caught this.

gaoharimran29-glitch (approved):
Approving this — the fixes from the last round all landed cleanly: CNAME chain error transparency, PTR concurrency, NSEC3 support, docstring updates, and test coverage for the previously-missing error paths are all in good shape and hold up under inspection.

Left two non-blocking comments. Doesn't need to hold up this PR, but feel free to open a quick follow-up whenever convenient to tighten that up. Nice work getting through all the rounds on this one.

PR #207, opened 2026-09-11
fix(helpers): try remaining candidates in list-typed date fields in safe_parse_datetime

### LINKED ISSUE #202: Bug: `safe_parse_datetime` only attempts the first element of a list-typed date field, with no fallback to remaining candidates
## Summary

`safe_parse_datetime` (`utils/helpers.py`) special-cases list input for WHOIS-style date fields (`expiration_date`, `creation_date`, `updated_date` can come back as `list[str]` from `whois_tool` depending on TLD/registrar). Per the existing comment in the code, this is intentional — it's there specifically to stop `str(list)` from being fed into the string-format parser, which would produce something like `"['2025-12-25 00:00:00']"` that matches no known format and silently suppresses downstream warnings.

However, the current handling only ever tries **`date_input[0]`**. If that first element is empty, malformed, or in a format none of the parser's format strings or the ISO fallback recognize, the function returns `None` immediately — it never attempts any subsequent element in the list, even if one of them would parse successfully. The net effect for a caller like `whois_extractor` is the same failure mode the original list-handling was added to prevent: a real, parseable date exists in the data, but the signal downstream (e.g. `domain_expiry_days`) ends up `None` with no indication that usable data was available and discarded.

## Severity

**Low–Medium.** This is a narrower edge case than "list vs string" — it depends on the *first* candidate specifically being unparseable while a later one isn't. Not something confirmed as currently happening with real registrar data, but plausible: registry-sourced and registrar-sourced date strings for the same domain aren't guaranteed to share a format, and ordering within the list isn't something `whois_tool` or its upstream (`python-whois`) documents as guaranteed. Given the whole point of the list-handling branch is to prevent exactly this class of silent signal loss, it's worth closing the gap rather than leaving a partial fix in place.

## Where it lives

- `utils/helpers.py` — `safe_parse_datetime`
- Consumed by: `tools/signals/whois.py` — `whois_extractor` (currently the only confirmed consumer of the list-unwrapping branch, via `expiration_date`)

## Relevant code path

```python
if isinstance(date_input, list):
    if not date_input:
        return None
    date_input = date_input[0]
    if not date_input:
        return None
    if isinstance(date_input, datetime):
        return date_input

clean_str = str(date_input).strip().replace("Z", "+00:00")

for fmt in ("%Y-%m-%dT%H:%M:%S%z", "%Y-%m-%d %H:%M:%S%z", "%Y-%m-%d %H:%M:%S", "%Y-%m-%d"):
    try:
        return datetime.strptime(clean_str, fmt)
    except (ValueError, TypeError):
        continue

try:
    return datetime.fromisoformat(clean_str)
except (ValueError, TypeError):
    return None
```

Once `date_input = date_input[0]` executes, every other element in the original list is discarded — win or lose, the function commits to that single candidate for the rest of its execution.

## Root cause

The list-unwrapping step reduces a `list[str]` to a single `str` *before* any parsing is attempted, rather than treating the list as an ordered set of candidates to try in sequence. This means "unwrap the list" and "parse a date string" are conflated into one step instead of being separable — there's no path back into the list if the chosen element turns out to be unparseable.

## Steps to reproduce (constructed — not yet observed with live registrar data)

1. Call `safe_parse_datetime(["not-a-real-date", "2028-09-14 04:00:00+00:00"])`.
2. Observe the function returns `None`, even though the second element is a perfectly valid, parseable date matching an existing format string.
3. Contrast with `safe_parse_datetime(["2028-09-14 04:00:00+00:00", "not-a-real-date"])`, which returns a valid `datetime` — the outcome depends entirely on list ordering rather than on whether *any* usable date exists in the input.

> **Note for anyone picking this up:** `full_recon` is currently unstable, so don't try to trigger or verify this through the full recon pipeline via MCP Inspector. This is testable in complete isolation — call `safe_parse_datetime` directly with constructed list inputs, no network calls or other tools required.

## Tests

Please add test coverage for this to:

```
tests\test_helpers.py
```

At minimum, cover:
- A list where the first element is unparseable but a later element is valid — confirm your fix's expected behavior (whether that's "falls back and succeeds" or some other deliberate, documented choice) with a real assertion, not just a lack of a crash.
- A list where every element is unparseable — should still cleanly return `None`.
- A list where the first element parses fine (today's already-working case) — confirm no regression.
- A single-element list, and an empty list — confirm existing behavior for these is preserved.

As with the earlier issue, these tests should exercise `safe_parse_datetime`  directly with constructed input — no dependency on `full_recon` or live network calls.

Files to be touched :- `tests\test_helpers.py` , `utils\helpers.py`

AnandkumarMall:
## Closes

Closes #202

## Summary

Fixes `safe_parse_datetime` to sequentially attempt each element in a list-typed date input until one parses successfully, rather than committing strictly to `date_input[0]` and returning `None` if the first element is empty, invalid, or malformed.

## What changed

- **`utils/helpers.py`**:
  - Replaced the single-element unwrapping `date_input = date_input[0]` with an iterative loop over `date_input`.
  - Attempts `safe_parse_datetime(item)` for each element in sequence; returns the first non-`None` parsed `datetime`.
  - Preserves immediate return for valid first elements and returns `None` only if all candidate elements fail to parse or the list is empty.
- **`tests/test_helpers.py`**:
  - Added test for single-element lists (`test_safe_parse_datetime_single_element_list`).
  - Added test verifying fallback when earlier elements are `None`, `""`, or unparseable strings, and subsequent elements contain valid dates (`test_safe_parse_datetime_list_fallback_when_earlier_elements_invalid`).
  - Added test confirming `None` is returned when all candidates in a list are unparseable (`test_safe_parse_datetime_list_all_elements_unparseable_returns_none`).
  - Maintained existing tests for single date strings, empty lists, and already-parsed datetime objects.

## Notes

- This ensures consumers such as `whois_extractor` do not lose `domain_expiry_days` telemetry when registrars return multiple expiration date records where the first candidate is empty or malformed.

## Verification

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -109,10 +109,29 @@ class TestHelpers(unittest.TestCase):
         """An empty list returns None, not a crash."""
         self.assertIsNone(safe_parse_datetime([]))
 
-    def test_safe_parse_datetime_list_with_empty_first_element(self):
-        """A list whose first element is empty/None falls through to None."""
-        self.assertIsNone(safe_parse_datetime([None, "2025-12-25"]))
-        self.assertIsNone(safe_parse_datetime(["", "2025-12-25"]))
+    def test_safe_parse_datetime_single_element_list(self):
+        """A single-element list parses correctly."""
+        result = safe_parse_datetime(["2025-12-25 00:00:00"])
+        self.assertEqual(result, datetime(2025, 12, 25, 0, 0, 0))
+
+    def test_safe_parse_datetime_list_fallback_when_earlier_elements_invalid(self):
+        """A list falls back to subsequent candidates when earlier elements are invalid, empty, or None."""
+        self.assertEqual(
+            safe_parse_datetime([None, "2025-12-25 00:00:00"]),
+            datetime(2025, 12, 25, 0, 0, 0),
+        )
+        self.assertEqual(
+            safe_parse_datetime(["", "2025-12-25 00:00:00"]),
+            datetime(2025, 12, 25, 0, 0, 0),
+        )
+        self.assertEqual(
+            safe_parse_datetime(["not-a-real-date", "2028-09-14 04:00:00+00:00"]),
+            datetime(2028, 9, 14, 4, 0, 0, tzinfo=timezone.utc),
+        )
+
+    def test_safe_parse_datetime_list_all_elements_unparseable_returns_none(self):
+        """A list where all candidates are unparseable returns None."""
+        self.assertIsNone(safe_parse_datetime([None, "", "not-a-date", "invalid-date-string"]))
 
     def test_safe_parse_datetime_invalid_string_returns_none(self):
         """A genuinely unparseable string returns None, not a crash."""
--- a/utils/helpers.py
+++ b/utils/helpers.py
@@ -74,18 +74,15 @@ def safe_parse_datetime(date_input) -> datetime | None:
         return date_input
 
     # python-whois returns a list of datetimes for some TLDs (e.g. when the
-    # registrar exposes both registry and registrar expiration dates). Take
-    # the first element — they are typically identical, and the alternative
-    # (str(list)) yields "['2025-12-25 00:00:00']" which matches no format
-    # and silently suppresses domain-expiry warnings downstream.
+    # registrar exposes both registry and registrar expiration dates). Try
+    # each candidate in sequence until one parses successfully, avoiding
+    # str(list) which yields "['...']" and matches no format.
     if isinstance(date_input, list):
-        if not date_input:
-            return None
-        date_input = date_input[0]
-        if not date_input:
-            return None
-        if isinstance(date_input, datetime):
-            return date_input
+        for item in date_input:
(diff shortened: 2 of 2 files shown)


gaoharimran29-glitch (approved):
Thanks for taking this on — the implementation is correct: recursing through safe_parse_datetime` per list element instead of hardcoding `date_input[0]` gets the fallback-to-remaining-candidates behavior we wanted, without duplicating any of the existing parsing logic.

Test coverage here is solid and, combined with the existing tests already in the file covering the first-element-valid case, satisfies the full checklist from the linked issue: fallback on `invalid`/`empty`/`None` elements, `all-elements-unparseable` returning `None`, and single-element-list behavior preserved. Nice attention to detail using full timestamps `("2025-12-25 00:00:00", "2028-09-14 04:00:00+00:00")` consistently across every test rather than bare dates — that exercises the `%Y-%m-%d %H:%M:%S` format branch specifically, which matches the shape of the real WHOIS data this is parsing, rather than only ever hitting the simpler date-only branch.

Merging this one.

PR #210, opened 2026-09-11
fix(helpers): try all WHOIS date candidates

### LINKED ISSUE #202: Bug: `safe_parse_datetime` only attempts the first element of a list-typed date field, with no fallback to remaining candidates

JavierQuinan:
Closes #202

## Summary

Treat list-valued WHOIS dates as ordered candidates instead of committing to the first element before parsing.

## What changed

- `safe_parse_datetime` now tries each list element in order and returns the first candidate that parses successfully;
- scalar strings and direct `datetime` inputs keep their existing behavior;
- empty lists and lists with no parseable values still return `None`;
- added direct unit coverage for:
  - invalid first candidate with a valid later candidate;
  - all-invalid candidate lists;
  - first-valid behavior;
  - single-element lists;
  - empty lists.

## Why

The previous implementation unwrapped `date_input[0]` before parsing. That meant a malformed or empty first value discarded later usable WHOIS dates and could suppress downstream expiry signals.

## Verification

The change is isolated to `utils/helpers.py` and `tests/test_helpers.py`, with no network dependency required for the new cases. The repository CI/test suite can provide the final execution validation for this PR.

### DIFF
--- a/tests/test_helpers.py
+++ b/tests/test_helpers.py
@@ -109,10 +109,26 @@ class TestHelpers(unittest.TestCase):
         """An empty list returns None, not a crash."""
         self.assertIsNone(safe_parse_datetime([]))
 
-    def test_safe_parse_datetime_list_with_empty_first_element(self):
-        """A list whose first element is empty/None falls through to None."""
-        self.assertIsNone(safe_parse_datetime([None, "2025-12-25"]))
-        self.assertIsNone(safe_parse_datetime(["", "2025-12-25"]))
+    def test_safe_parse_datetime_list_falls_back_to_later_valid_element(self):
+        """A later valid candidate is used when earlier values cannot be parsed."""
+        expected = datetime(2028, 9, 14, 4, 0, 0, tzinfo=timezone.utc)
+        self.assertEqual(
+            safe_parse_datetime(["not-a-real-date", "2028-09-14 04:00:00+00:00"]),
+            expected,
+        )
+        self.assertEqual(safe_parse_datetime([None, "2025-12-25"]), datetime(2025, 12, 25))
+        self.assertEqual(safe_parse_datetime(["", "2025-12-25"]), datetime(2025, 12, 25))
+
+    def test_safe_parse_datetime_list_all_invalid_returns_none(self):
+        """A list with no parseable candidates still returns None."""
+        self.assertIsNone(safe_parse_datetime(["not-a-date", "", None, "also-not-a-date"]))
+
+    def test_safe_parse_datetime_single_element_list(self):
+        """A single-element list preserves the existing successful behavior."""
+        self.assertEqual(
+            safe_parse_datetime(["2025-12-25"]),
+            datetime(2025, 12, 25),
+        )
 
     def test_safe_parse_datetime_invalid_string_returns_none(self):
         """A genuinely unparseable string returns None, not a crash."""
--- a/utils/helpers.py
+++ b/utils/helpers.py
@@ -73,19 +73,16 @@ def safe_parse_datetime(date_input) -> datetime | None:
     if isinstance(date_input, datetime):
         return date_input
 
-    # python-whois returns a list of datetimes for some TLDs (e.g. when the
-    # registrar exposes both registry and registrar expiration dates). Take
-    # the first element — they are typically identical, and the alternative
-    # (str(list)) yields "['2025-12-25 00:00:00']" which matches no format
-    # and silently suppresses domain-expiry warnings downstream.
+    # python-whois can return multiple candidate dates for some TLDs
+    # (for example, registry and registrar expiration dates). Try candidates
+    # in order and return the first value that parses successfully rather than
+    # discarding later usable values when the first candidate is malformed.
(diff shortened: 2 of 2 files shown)


gaoharimran29-glitch (commented):
Thanks for the PR — this is functionally correct and lines up with the fix we're merging elsewhere (recursive fallback through the list rather than only checking the first element). Your `all_invalid_returns_none` test is a good one too — it already mixes `None`, `""`, and invalid strings in a single case, which is exactly the coverage this needed.

Three of us converged on the same implementation independently for this issue, so we're consolidating on a single merged PR (#207) rather than carrying duplicate logic across branches.

One thing worth adopting for next time: most of the dates here are bare (`"2025-12-25"`), which only exercises the `%Y-%m-%d` parsing branch — worth using full timestamps like `"2025-12-25 00:00:00"` instead, since that's closer to the actual shape of WHOIS data this function handles in production.

Appreciate the quick turnaround on this.

PR #213, opened 2026-09-13
feat(email-security): harden SPF/DKIM/DMARC record validation

### LINKED ISSUE #152: Improvement :-  SPF/DKIM/DMARC record validation + minor cleanup in email_security_check

LouisDeconinck:
## Closes

Closes #152

## Summary

Improves the accuracy of SPF, DKIM, and DMARC validation in `email_security_check` — fixes false positives/negatives, detects ambiguous multi-record configurations, and performs the requested minor cleanup. Tag-aware parsing follows RFC 6376 (DKIM), RFC 7208 (SPF), and RFC 7489/9989 (DMARC).

## What changed

- **SPF**
  - New `_is_spf_record()`: only a first token of exactly `v=spf1` qualifies (RFC 7208 §4.5) — `V=SPF1`, `v=spf1foo`, and `v=spf1;-all` are discarded instead of parsed.
  - `_spf_policy()` now tokenizes the record and evaluates mechanisms left-to-right (RFC 7208 §4.6.2): the first `all` mechanism wins; `-all`/`~all`/`?all`/`+all`/bare `all` all map correctly; substrings like `-allow` or `include:send-all.example.com` no longer false-match.
  - Distinct outcomes: `redirect` (redirect= present, no `all` — RFC 7208 §6.1), `missing` (no `all`/`redirect=`), `unknown` (not an SPF record).
  - Multiple SPF records → `found: true, valid: false, record: null`, all candidates preserved in `records`, with a consolidation recommendation.
- **DKIM**
  - New `_parse_dkim_record()` classifies each TXT record as `active` / `revoked` / `malformed` / `not_dkim` per RFC 6376 (link in docstring): case-sensitive tag names matching `[A-Za-z][A-Za-z0-9_]*` (extension tags like `x_foo` accepted), no whitespace inside tag names, duplicate tags invalidate, `v=` is optional but must be the first tag and exactly `DKIM1`, and `p=` must be non-empty base64 key data.
  - `_check_dkim` now reports `found_selectors`, `revoked_selectors`, and `malformed_selectors` separately; revoked keys (`p=` empty) are no longer counted as active.
  - The "not found" recommendation now notes DKIM may still exist under an unchecked selector (best-effort honesty).
- **DMARC**
  - Multiple DMARC records → `valid: false`, all preserved, single-record recommendation.
  - `rua` missing vs present-but-empty are distinguished; empty `rua=` is treated as unconfigured (-5).
  - `p=` validated against `none`/`quarantine`/`reject` → `policy: "invalid", valid: false` (optional item 5a); missing `p=` defaults to `none` per RFC 9989 §4.7 with a recommendation.
- **Cleanup**: removed redundant `domain.lower()` (normalize_domain owns casing); moved `import datetime` to top-level imports; deterministic sorted selector list.
- **Docs**: updated `docs/tools/email_security_check.md` for the new fields (`valid`, `records`, `revoked_selectors`, `malformed_selectors`), policy values, and scoring.

## Notes

- Aware of the existing open PRs for this issue (#153, #205). Per the maintainer's comment on #152 inviting new PRs, this is an alternative implementation. Where applicable it already incorporates the review feedback raised on #205 (case-sensitive `v=DKIM1` as first tag, duplicate-tag rejection, `v =DKIM1` whitespace rejection, base64 validation of `p=`, left-to-right `all` evaluation, redirect= handling, `record: null` on multi-record results, missing-`all` scored below neutral, `malformed_selectors` collection, RFC reference links).
- Scoring is unchanged for the standard cases (fail=30, softfail=20, neutral/pass=10, DMARC reject=35/quarantine=25, DKIM active=35); new states: redirect=25, missing-all=5, unknown=0.

## Verification

- [x] Ran locally and confirmed it works as expected — live check against google.com correctly identifies its revoked DKIM selectors (`20161025`, `20230601` publish `p=` empty) which the old code reported as found; github.com/outlook.com selectors report `active`.
- [x] Added/updated tests for the changed behavior (19 tests in `tests/test_email_security.py` covering tokenized SPF policies, multi-record SPF/DMARC, DKIM active/revoked/malformed/not-DKIM, empty `rua=`, invalid `p=`, missing `p=`)
- [x] All tests pass (`uv run pytest tests/` — 561 passed)
- [x] Updated Documentation (docs/tools/email_security_check.md)


docs/tools/email_security_check.md
```diff
@@ -103,36 +114,47 @@ top-level fields:
 
 | Field | Type | Description |
 |---|---|---|
-| `found` | boolean | Whether a TXT record beginning with `v=spf1` was found. |
-| `record` | string or null | The first matching SPF record. |
-| `policy` | string or null | `fail` for `-all`, `softfail` for `~all`, `neutral` for `?all`, `pass` for `+all`, `unknown` when none of those mechanisms is present, or `null` when no SPF record was found. |
+| `found` | boolean | Whether at least one TXT record starting with an exact `v=spf1` version token was found. |
+| `valid` | boolean | `false` when multiple SPF records exist or the record could not be parsed. |
```
gaoharimran29-glitch, line 118: Nit (non-blocking): the SPF valid field documentation says it becomes `false` when a record cannot be parsed. In the current implementation, records failing the SPF version-token check are filtered out and reported as `found: false`; `valid: false` is primarily returned for multi-record ambiguity. Please tighten this wording for exact behavioral parity with `_is_spf_record()` / `_check_spf()`.

gaoharimran29-glitch (approved):
Approved. This implementation satisfies the #152 requirements and resolves the outstanding review concerns around SPF ambiguity handling, DKIM record classification, malformed/revoked selector visibility, and response-schema consistency. The regression coverage is appropriately targeted, and the documentation accurately reflects the expanded result contract and scoring behavior. Nice work keeping the changes focused and test-backed.

gaoharimran29-glitch:
This PR was reviewed and approved, and is the implementation we’ll proceed with for this issue.

There was an accidental merge of another independent PR addressing the same issue while preparing to merge this one. That merge has now been reverted, restoring the target branch to the expected state.

I’ve verified that this PR is currently mergeable without conflicts, so we can proceed with the intended implementation here.

Thanks for the contribution and for the clean implementation.

PR #214, opened 2026-09-13
feat(takeover): return structured probe evidence and confidence for findings

LouisDeconinck:
## Closes

Addresses #154 (Item 11)

## Summary

Confirmed takeover findings now carry structured probe evidence (requested URL, final URL, status code, redirect chain, matched indicator) and a `confidence` rating reported separately from `severity`, so a fingerprint match is no longer presented as an equally-confirmed exploit regardless of how it was observed.

## What changed

- **`tools/subdomain_takeover_tool.py`**:
  - `_probe()` records the request URL that produced a response and includes the attempted URL in each per-scheme probe error.
  - `_confirms_takeover()` builds an `evidence` payload for every completed probe: `url`, `final_url`, `status_code`, `redirect_chain`, `cross_host_redirect`, and `matched_indicator` (`{"type": "body"|"status", "value": ...}`, or `null` when nothing matched).
  - Added `_response_url()` / `_redirect_chain()` helpers that reconstruct the followed redirect chain from `response.history`.
  - Added `_assess_confidence()`: a service-specific body marker served by the probed host rates `high`; a bare status-code match or a match after a redirect to a different host rates `medium`; both together rate `low`.
  - Vulnerable findings gain `confidence` and `evidence` keys alongside the existing `severity`.
- **`tests/test_subdomain_takeover.py`**:
  - Evidence fields on a confirmed finding (URL, final URL, status code, matched indicator, redirect chain).
  - Cross-host redirect is recorded and downgrades confidence to `medium`.
  - Status-only indicator match yields `medium`, and `low` after a cross-host redirect.
  - A no-indicator probe still returns its observed evidence.
  - Failed probes keep the attempted URL alongside each transport error.

## Notes

- `severity` (impact if exploited) is intentionally unchanged; `confidence` describes how conclusive the evidence is.
- The flat `not_vulnerable` list and all existing result keys keep their current shape for schema compatibility; evidence is attached where findings previously lacked it.
- Scoped to item 11 only — items 5–9 and 12 are covered by #208.

## Verification

- [x] All tests pass (`uv run pytest tests/test_subdomain_takeover.py` — 29 passed)


gaoharimran29-glitch:
Thanks for the contribution! Since the related PR (#208) has now been merged, this branch is showing conflicts with `main`.

Please update your branch with the latest `main` and resolve the conflicts while preserving your changes. Once the conflicts are resolved and CI passes, I'll start review it.

LouisDeconinck:
Rebased onto main after #208. The probe's redirect handling changed upstream to manual per-hop following (SSRF check + same-subdomain scope guard per hop), so this PR's evidence fields were merged onto that: `redirect_chain` now reports the URLs the manual follower actually requested (rather than `response.history`, which is empty under `allow_redirects=False`), and probe errors keep their `url`. One consequence worth noting: out-of-scope redirect targets are never fetched anymore, so `cross_host_redirect` can only trigger on redirects that stay within the scanned subdomain's scope. Tests updated accordingly; suite is green (58 takeover tests, 590 overall).

gaoharimran29-glitch (changes requested):
Could you address the external-redirect handling highlighted by Copilot?

When `_probe()` encounters a redirect outside the target subdomain scope, we intentionally stop following it, but the rejected `Location` is currently not preserved. As a result, `final_url` / `cross_host_redirect` can incorrectly reflect the original URL.

Please preserve the rejected redirect target in the probe metadata without fetching it, and use it when constructing the redirect chain, `final_url`, and confidence/evidence fields. Please also add regression coverage for this case, including verifying that the external target is not requested.

For the `matched_indicator: "headers"` concern, please first verify whether there is an existing evidence schema/type that restricts the indicator type to `body` / `status`. The current implementation supports `headers`, but all current fingerprints also contain a body, so this branch is not reached with the existing fingerprints. If an external schema does restrict it, please update the schema consistently; otherwise no change is needed there.

LouisDeconinck:
Addressed the rejected-redirect evidence gap.

`_probe()` now records the out-of-scope `Location` (resolved to an absolute URL via `urljoin`) on a new `rejected_redirect_url` field of `_ProbeResult` when the scope guard halts following — the target is preserved as metadata only and is never requested, so the scope/SSRF guard is not weakened. `_confirms_takeover()` then consumes it: the rejected target is appended as the terminal hop of the evidence `redirect_chain`, reported as `final_url`, and exposed via a `rejected_redirect_url` evidence key so the unfetched hop cannot be mistaken for a fetched one. Because `final_url` now reflects the external target, `cross_host_redirect` triggers on rejected redirects and `_assess_confidence` downgrades accordingly (body match → `medium`, status-only → `low`).

Regression coverage:
- `test_probe_preserves_rejected_out_of_scope_redirect` / `test_probe_preserves_rejected_redirect_after_in_scope_hops`: rejected target lands in `rejected_redirect_url` while `redirect_chain` stays limited to actually-requested URLs, and the mocked `requests.get` is asserted (call count + per-call URLs) to never have been called with the external URL.
- `test_rejected_redirect_becomes_terminal_hop_in_evidence` / `test_body_match_with_rejected_redirect_is_medium_confidence`: evidence chain ends at the external target, `final_url` reports it, `cross_host_redirect` is `True`, confidence drops to `low`/`medium`, and the external URL is never fetched.
- The existing structured-evidence test now also asserts `rejected_redirect_url` is `None` on a normal finding.

On the Copilot `matched_indicator: "headers"` finding — verified: there is no evidence schema/type restricting the indicator type to `body`/`status`. `matched_indicator` is built only in `_confirms_takeover` and is not deserialized by any consumer (the tool is not wired into the signals registry/full_recon pipeline, and there is no TypedDict/JSON schema for the evidence payload — the `body|status` mention is only in the PR description). All current `VULNERABLE_FINGERPRINTS` define `status` + `body` + `headers`, so the `elif "headers"` branch is unreachable today and `body` is always emitted; it stays as the correct fallback for a hypothetical headers-only fingerprint. No change needed there.

Suite is green: 63 takeover tests, 600 overall.

gaoharimran29-glitch (approved):
LGTM. The external redirect handling now preserves the rejected `Location` without following it, and the redirect metadata is correctly propagated into `redirect_chain`, `final_url`, and `cross_host_redirect` for confidence assessment.

The regression coverage also addresses the relevant redirect behavior.

Thanks for addressing the review feedback. Approving.

